PR #76, opened 2017-11-15
Azure blob storage

kenneth-hendricks:


README.md
```diff
@@ -50,14 +51,22 @@ https://github.com/catalyst/moodle-local_datacleaner
...
 2. Setup your remote object storage. See [Remote object storage setup](#remote-object-storage-setup)
 3. Clone this repository into admin/tool/objectfs
-4. Clone [moodle-local_aws](https://github.com/catalyst/moodle-local_aws) into local/aws
-4. Install the plugins through the moodle GUI.
-5. Configure the plugin. See [Moodle configuration](#moodle-configuration)
-6. Place the following line inside your Moodle config.php:
-
+4. Install one of the required SDK libraries for the storage file system that you will be using
+    1. Clone [moodle-local_aws](https://github.com/catalyst/moodle-local_aws) into local/aws, or
+    2. Clone [moodle-local_azure](https://github.com/catalyst/moodle-local_azure) into local/azure
```
kenneth-hendricks: broken link https://github.com/catalyst/moodle-local_azure_storage

classes/client/azure_storage_client.php
```diff
@@ -0,0 +1,362 @@
...
+        $mform->setType("accountname", PARAM_TEXT);
+
+        $mform->addElement('text', 'container', get_string('settings:azure:container', 'tool_objectfs'));
+        $mform->addHelpButton('container', 'settings:azure:container', 'tool_objectfs');
+        $mform->setType("container", PARAM_TEXT);
+
+        $mform->addElement('text', 'sastoken', get_string('settings:azure:sastoken', 'tool_objectfs'));
+        $mform->addHelpButton('sastoken', 'settings:azure:sastoken', 'tool_objectfs');
+        $mform->setType("sastoken", PARAM_RAW);
+
```
kenneth-hendricks: Help text blank for these 3

README.md
```diff
@@ -4,8 +4,7 @@
 
 # moodle-tool_objectfs
```
kenneth-hendricks: Now that we have two storage options may be wise to have a switching object storage section. Should have steps to follow to avoid issues. roughly:

1. Pull all remote objects to duplicated
2. truncate object table and object report table
3. switch

kenneth-hendricks: Maybe place in a FAQ section that we can add to as needed

README.md
```diff
@@ -15,10 +14,12 @@ A remote object storage file system for Moodle. Intended to provide a plug-in th
 * [Currently supported object stores](#currently-supported-object-stores)
   * [Roadmap](#roadmap)
   * [Amazon S3](#amazon-s3)
+  * [Azure Blob Storage](#azure-blob-storage)
 * [Moodle configuration](#moodle-configuration)
   * [General Settings](#general-settings)
   * [File Transfer settings](#file-transfer-settings)
   * [Amazon S3 settings](#amazon-s3-settings)
+  * [Azure Blob Storage settings](#azure-blob-storage-settings)
```
kenneth-hendricks: broken link

kenneth-hendricks:
Basic testing went well confirmed:
* Embeded audio
* Embeded video
* Basic docs

could all be served from azure correctly

this also confirms that the file handle provided is fseekable.

classes/object_manipulator/pusher.php
```diff
@@ -48,6 +48,13 @@ class pusher extends manipulator {
     private $minimumage;
 
     /**
+     * The maximum upload file size in bytes.
```
kenneth-hendricks: My bad for leaving them in here but do you mind removing the 'use Aws\S3\Exception\S3Exception;' statments from the puller, pusher, recoverer and deleter?

lib.php
```diff
@@ -88,6 +84,20 @@ function get_objectfs_config() {
     $config->logging = 0;
     $config->preferexternal = 0;
 
+    // TODO: load based on fs? move to client files?
+
+    // '\tool_objectfs\s3_file_system'
```
kenneth-hendricks: Thoughts on splitting the definition of these object storage specific config variables out into a seperate object_settings (or another name) class . We could also move test_connection, test_permissions and define_client_section into this class to keep the object client classes more focused

kenneth-hendricks: I would also namespace the config variable names to avoid collisions in the future. 

e.g. $config->aws_key

though this namespacing could be handled outside of the new object settings class if implemented.

tests/object_file_system_test.php
```diff
@@ -353,5 +353,17 @@ public function test_remove_file_will_not_remove_remote_file() {
         $isremotereadable = $this->is_externally_readable_by_hash($filehash);
         $this->assertTrue($isremotereadable);
     }
+
+    public function test_file_is_seekable() {
```
kenneth-hendricks: nice, been meaning to add something like this

classes/form/settings_form.php
```diff
@@ -111,66 +110,24 @@ public function define_file_transfer_section($mform, $config) {
...
-        $connection = false;
-
-        $client = new s3_client($config);
-        $connection = $client->test_connection();
+    public function define_client_section($mform, $config) {
+        global $CFG, $OUTPUT;
 
-        if ($connection->success) {
-            $mform->addElement('html', $OUTPUT->notification($connection->message, 'notifysuccess'));
+        if (empty($CFG->alternative_file_system_class)) {
```
kenneth-hendricks: We're lose the ability to pre-set client config before we flip on the alternative_file_system_class - do you think this is important enough to change how we handle this ?

Couple alternatives i can think of:

1. Allow client to set which object storage they want to use as a setting. We can then also check that alternative_file_system_class matches the expected class for that client and warn if it doesnt.
2. Just show all client sections here if alternative_file_system_class isn't set with appropriate text that alternative_file_system_class isnt to their respective classes in each section

db/upgrade.php
```diff
@@ -42,5 +42,31 @@ function xmldb_tool_objectfs_upgrade($oldversion) {
...
+        $config = get_config('tool_objectfs');
+
+        // Remapping the new namespaced variables.
+        set_config('s3_key', $config->key, 'tool_objectfs');
+        set_config('s3_secret', $config->secret, 'tool_objectfs');
+        set_config('s3_bucket', $config->bucket, 'tool_objectfs');
+        set_config('s3_region', $config->region, 'tool_objectfs');
+
+        // Use the existing filesystem that was once hardcoded.
+        set_config('filesystem', '\\tool_objectfs\\azure_file_system', 'tool_objectfs');
```
kenneth-hendricks: Shouldnt this be s3_file_system ? all upgrading installs will be on that

PR #92, opened 2018-05-01
Added all Amazon regions as of 1st May 2018

jackson-catalyst:
We were missing eu-west-2 region so I have added it to the list. 

I have also added in all the regions as per https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html


kenneth-hendricks:
Seems to have been taken care of in https://github.com/catalyst/moodle-tool_objectfs/commit/7a09fe95cf2810990474467ea686110bbdded296 .

Closing

Issue #97, opened 2018-06-06
Detailed report for Missing from filedir and external storage


dmitriim:
1. Add a link to a new report - Missing from filedir and external storage ( see details )
2. A new page with the list of missing files

kenneth-hendricks:
Closing as merged

PR #98, opened 2018-06-07
Add new report for missing files to close issue #97

dmitriim:


classes/table/files_table.php
```diff
@@ -0,0 +1,101 @@
...
+ * @author    Dmitrii Metelkin <dmitriim@catalyst-au.net>
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace tool_objectfs\table;
+
+defined('MOODLE_INTERNAL') || die();
+
+class files_table extends \table_sql {
```
kenneth-hendricks: Getting 

```
moodle_1          | 2018/06/07 02:56:36 [error] 37#37: *206 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Class 'table_sql' not found in /siteroot/admin/tool/objectfs/classes/table/files_table.php on line 30
moodle_1          | PHP message: PHP Stack trace:
moodle_1          | PHP message: PHP   1. {main}() /siteroot/admin/tool/objectfs/missing_files.php:0
moodle_1          | PHP message: PHP   2. spl_autoload_call() /siteroot/admin/tool/objectfs/missing_files.php:37
moodle_1          | PHP message: PHP   3. core_component::classloader() /siteroot/admin/tool/objectfs/missing_files.php:37
```

On mdl34 - what version were you running this on ?

Adding 'require_once($CFG->libdir . '/tablelib.php'); ' at the top fixed this

classes/table/files_table.php
```diff
@@ -0,0 +1,101 @@
...
+
+    public function col_contextid(\stdClass $row) {
+        return $row->contextid;
+    }
+
+    public function col_contenthash(\stdClass $row) {
+        return $row->contenthash;
+    }
+
+    public function col_path(\stdClass $row) {
```
kenneth-hendricks: Don't see the point of this column - also if included it should really be local file path as there's no guarantee that it will be stored in this format on all external file systems. 

mdl_files.filepath does have meaning though, i would show that here instead.

classes/table/files_table.php
dmitriim: Works ok on 3.1

classes/table/files_table.php
dmitriim: That was a client requirement. I'll rename to "local file path "

PR #100, opened 2018-06-15
Privacy tests

ilya-catalyst:


marcusboon:
You might want to squash those commits into one and then change the commit message as it refers to datacleaner and this is objectfs.

PR #103, opened 2018-06-25
Wr#290488

ilya-catalyst:


classes/s3_file_system.php
```diff
@@ -40,4 +43,26 @@ protected function get_external_client($config) {
...
+
+    public function copy_object_from_local_to_external_path($contenthash) {
+
+        $config = get_objectfs_config();
+        $s3client = $this->get_external_client($config);
+
+        $localpath = $this->get_local_path_from_hash($contenthash);
+        $externalpath = $s3client->get_filepath_from_hash($contenthash);
+
+        $uploader = new MultipartUploader($s3client->get_client(), $localpath, [
```
kenneth-hendricks: All direct s3 logic should inside s3_client  itself. Would wrap this in a s3_client method such that code looks like:

$s3client->copy_object_from_local_to_external_path($localpath, $contenthash); 

get_filepath_from_hash can also remain private this way.

classes/client/s3_client.php
```diff
@@ -87,12 +87,16 @@ public function set_client($config) {
         ));
     }
 
+    public function get_client() {
```
kenneth-hendricks: shouldnt exist, this is a wrapper class

classes/client/s3_client.php
```diff
@@ -150,7 +150,7 @@ public function get_seekable_stream_context() {
         return $context;
     }
 
-    protected function get_filepath_from_hash($contenthash) {
+    public function get_filepath_from_hash($contenthash) {
```
kenneth-hendricks: This needs to remain protected

classes/client/s3_client.php
```diff
@@ -333,4 +333,22 @@ public function define_client_section($mform, $config) {
...
+        $externalpath = $this->get_filepath_from_hash($contenthash);
+
+        $uploader = new MultipartUploader($this->client, $localpath, [
+            'bucket' => $this->bucket,
+            'key'    => $externalpath,
+        ]);
+
+        try {
+            $result = $uploader->upload();
+            echo "Upload complete: {$result['ObjectURL']}\n";
```
kenneth-hendricks: Dont need this echo,

classes/s3_file_system.php
```diff
@@ -40,4 +40,11 @@ protected function get_external_client($config) {
         $s3client = new s3_client($config);
         return $s3client;
     }
+
+    public function copy_object_from_local_to_external_path($contenthash) {
+        $config = get_objectfs_config();
+        $s3client = $this->get_external_client($config);
+        $localpath = $this->get_local_path_from_hash($contenthash);
+        $s3client->copy_object_from_local_to_external_path($localpath, $contenthash);
```
kenneth-hendricks: should 

return $s3client->copy_object_from_local_to_external_path($localpath, $contenthash);

classes/client/s3_client.php
```diff
@@ -333,4 +333,22 @@ public function define_client_section($mform, $config) {
...
+            return $result;
+        } catch (MultipartUploadException $e) {
+            echo $e->getMessage() . "\n";
+        }
+    }
```
kenneth-hendricks: Neet to return true on success and false if anything fails to match the api of the php copy() function

kenneth-hendricks:
Something fishy is going on, getting an infinite download with 6, text/plain GB file.

I think being text/plain is causing it to not use byteserving_send_file, instead using readfile_allow_large which may not work for large files in S3. 

Will investigate more when i have more time, this is a blocker.

Issue #108, opened 2018-09-27
Performance - SQL redundant condition?


egiles:
Some of the SQL also have MAX(filesize). 

Why is this there when file sizes should always match the contenthash? To make it easier to find collisions? If not its additional processing (sorry to split hairs).

mattporritt:
Marking this as nofix.
There are cases where there are differences so the max is required

dmitriim:
I'd like to reopen this issue as this actually have a great impact on the systems with a huge number of files in files tables. See DB average CPU usage before and after enabling tasks with these queries. It's like +20% to av CPU usage. 

![image](https://user-images.githubusercontent.com/4626431/58914673-7d492080-8762-11e9-93cb-06f2a4d22832.png)

TrueBurn:
This one query accounts for almost half the load on average on my entire LMS database server:

![image](https://user-images.githubusercontent.com/38687879/175030191-4fac9ddb-3e8c-40f4-bd50-eb6ea6ab26fb.png)

Query plan shows massive table scans even when nothing is returned and takes over 4 seconds to run:

![image](https://user-images.githubusercontent.com/38687879/175030453-1d50bc23-b35f-494d-a9e9-802153497a00.png)

PR #115, opened 2018-10-15
Issue #108 - remove redundant Max() in 'get candidate object' query

tuanngocnguyen:
As suggested by Evan https://github.com/catalyst/moodle-tool_objectfs/issues/108.

### DIFF
--- a/classes/object_manipulator/deleter.php
+++ b/classes/object_manipulator/deleter.php
@@ -87,15 +87,15 @@ class deleter extends manipulator {
         }
 
         $sql = 'SELECT f.contenthash,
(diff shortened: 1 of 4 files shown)


dmitriim:
@tuanngocnguyen  travis is very very unhappy

kenneth-hendricks:
Just gonna chime in here - seen alot of moodles where there are file records with same contenthash and different filesizes - thats why the max was originally here.

mattporritt:
Based on KC’s comment if there are cases where there are differences, which I think I’ve seen to then we should not remove the max statement

Issue #93, opened 2018-05-26
Misconfigured ObjectFS plugin causing cron job to fail


hosungsmsft:
In https://github.com/Azure/Moodle/issues/133, it is discovered that a misconfigured ObjectFS plugin causes the Moodle cron job to constantly fail due to `!!! Exception - Class '\tool_objectfs\azure_file_system' not found !!!`. I confirm the same myself (for Azure Blob Storage option, with a blank SAS token), and I think a misconfigured ObjectFS plugin shouldn't affect the Moodle cron job like this. Could you please check and confirm if this is the current behavior, and fix this issue if so?

PR #117, opened 2018-10-30
cron job fail when Objectfs is misconfigured Issue#93

tuanngocnguyen:
issue #93 - cron job fail when Objectfs is misconfigured


classes/object_manipulator/manipulator.php
```diff
@@ -133,10 +133,10 @@ public static function setup_and_run_object_manipulator($manipulatorclassname) {
 
         $shouldtaskrun = tool_objectfs_should_tasks_run();
 
-        if ($shouldtaskrun) {
```
kenneth-hendricks: Would display a different message in the case that the client is not available. Also lets use some early exits to avoid nesting hell like so:

```
diff --git a/classes/object_manipulator/manipulator.php b/classes/object_manipulator/manipulator.php
index 1799e08..9214a8c 100644
--- a/classes/object_manipulator/manipulator.php
+++ b/classes/object_manipulator/manipulator.php
@@ -131,18 +131,20 @@ abstract class manipulator {
     public static function setup_and_run_object_manipulator($manipulatorclassname) {
         $config = get_objectfs_config();
 
-        $shouldtaskrun = tool_objectfs_should_tasks_run();
-
         $filesystem = new $config->filesystem();
 
-        if ($shouldtaskrun && $filesystem->get_client_availability()) {
-            $logger = new \tool_objectfs\log\aggregate_logger();
-
-            $manipulator = new $manipulatorclassname($filesystem, $config, $logger);
-            $candidatehashes = $manipulator->get_candidate_objects();
-            $manipulator->execute($candidatehashes);
-        } else {
+        if (!tool_objectfs_should_tasks_run()) {
             mtrace(get_string('not_enabled', 'tool_objectfs'));
+            return;
+        }
+
+        if (!$filesystem->get_client_availability()) {
+            mtrace(get_string('client_not_available', 'tool_objectfs'));
+            return;
         }
+        $logger = new \tool_objectfs\log\aggregate_logger();
+        $manipulator = new $manipulatorclassname($filesystem, $config, $logger);
+        $candidatehashes = $manipulator->get_candidate_objects();
+        $manipulator->execute($candidatehashes);
     }
 }
diff --git a/lang/en/tool_objectfs.php b/lang/en/tool_objectfs.php
index f61ddb7..752f68b 100644
--- a/lang/en/tool_objectfs.php
+++ b/lang/en/tool_objectfs.php
@@ -33,6 +33,7 @@ $string['recover_error_objects_task'] = 'Object error recovery task';
 
 $string['generate_status_report_task'] = 'Object status report generator task';
 $string['not_enabled'] = 'The object file system background tasks are not enabled. No objects will move location until you do.';
+$string['client_not_available'] = 'The configured remote client is not available. Please ensure it is installed correctly.';
 
 $string['object_status:page'] = 'Object status';
 $string['object_status:location'] = 'Object location';
```

Issue #99, opened 2018-06-11
Moodle proxy setting is ignored


cristianomariacumer:
Hello,

in seems that objectfs is ignoring the proxy setting of the moodle installation. This is problematic if moodle is hosted in a DMZ where egress traffic is not permitted

brendanheywood:
Tasks

* [ ] Have a new admin setting which say wether or not to use the moodle proxy
* [ ] If true the push the moodle proxy settings into guzzle (we could also use the moodle curl lib but I think that is way too much work)

PR #118, opened 2018-11-01
Add proxy setting for aws Issue#99

tuanngocnguyen:
Add Proxy Setting for AWS #99


classes/client/s3_client.php
```diff
@@ -80,10 +80,34 @@ public function __wakeup() {
...
+            }
+        }
+
         $this->client = S3Client::factory(array(
         'credentials' => array('key' => $config->s3_key, 'secret' => $config->s3_secret),
         'region' => $config->s3_region,
-        'version' => AWS_API_VERSION
+        'version' => AWS_API_VERSION,
+            'http'    => [
+                'proxy' => $proxy,
```
dmitriim: would be good to abstract the code on lines 83-103 to a function or maybe even multiple functions. At the moment it's really hard to follow what we are expecting. There are 3 levels of conditions. Would be good somehow to cover it by tests.

dmitriim: Also would it be safe to pass an empty proxy? would it be the same as no proxy?

classes/client/s3_client.php
```diff
@@ -331,6 +355,33 @@ public function define_client_section($mform, $config) {
 
         $mform->addElement('select', 's3_region', get_string('settings:aws:region', 'tool_objectfs'), $regionoptions);
         $mform->addHelpButton('s3_region', 'settings:aws:region', 'tool_objectfs');
+
+        $mform = $this->define_proxy_config($mform, $config);
+
+        return $mform;
+    }
+
+    private function define_proxy_config($mform, $config) {
```
dmitriim: Like this abstraction :+1:

classes/client/s3_client.php
```diff
@@ -331,6 +355,33 @@ public function define_client_section($mform, $config) {
 
         $mform->addElement('select', 's3_region', get_string('settings:aws:region', 'tool_objectfs'), $regionoptions);
         $mform->addHelpButton('s3_region', 'settings:aws:region', 'tool_objectfs');
+
+        $mform = $this->define_proxy_config($mform, $config);
+
+        return $mform;
```
dmitriim: why do we return the form here?

classes/client/s3_client.php
tuanngocnguyen: Yep, I will try to abstract it and create a test.

tuanngocnguyen: I have tested it in the UI, with work as "no proxy". However, it will put a condition there as it will be safer.

classes/client/s3_client.php
```diff
@@ -80,11 +80,21 @@ public function __wakeup() {
     }
 
     public function set_client($config) {
-        $this->client = S3Client::factory(array(
-        'credentials' => array('key' => $config->s3_key, 'secret' => $config->s3_secret),
-        'region' => $config->s3_region,
-        'version' => AWS_API_VERSION
-        ));
+        $proxy = $this->construct_proxy_string($config);
```
dmitriim: Hi @tuanngocnguyen  It's still a lot of code duplication. What do you think about abstracting the array we are passing to S3Client::factory() to a function? There we can check for proxy and add additional key if required...

classes/client/s3_client.php
```diff
@@ -331,6 +341,55 @@ public function define_client_section($mform, $config) {
...
+        $mform->addElement('password', 'proxypassword', get_string('settings:proxypassword', 'tool_objectfs'));
+        $mform->addHelpButton('proxypassword', 'settings:proxypassword', 'tool_objectfs');
+        $mform->setType("proxypassword", PARAM_TEXT);
+    }
+
+    private function construct_proxy_string($config) {
+        $proxy = '';
+        if (!empty($config->proxyserver)) {
+            if (!empty($config->proxyusername)) {
+                $arrayproxy = explode("//", $config->proxyserver);
```
dmitriim: Would be good to add some comments here to understand what could com as $config->proxyserver and explain why we have som complicated code below. Test are appreciated as well ;)

classes/client/s3_client.php
brendanheywood: Isn't the point of this to honor the pre existing moodle proxy config which already has it's own config here:

https://github.com/moodle/moodle/blob/master/admin/settings/server.php#L102-L112

Generally speaking if we use the built in moodle curl library you get this for free:

https://github.com/moodle/moodle/blob/master/lib/filelib.php#L2946

But I suspect this will not be the case because we use GuzzleHttp but it would be case by case depending on S3 vs azure etc?

Either way this whole patch will need a rebase / rework against the dev branch which has moved on a fair bit

classes/client/s3_client.php
```diff
@@ -335,9 +345,60 @@ public function define_client_section($mform, $config) {
 
         $mform->addElement('select', 's3_region', get_string('settings:aws:region', 'tool_objectfs'), $regionoptions);
         $mform->addHelpButton('s3_region', 'settings:aws:region', 'tool_objectfs');
+
```
brendanheywood: A large patch has been merged which converts all the settings from formslib to admin settings, this will need to be rebased and this section converted - BUT as I said above, I don't think this patch should add any settings, it should honor the existing moodle proxy settings

brendanheywood:
Closing this, it's landed in 

https://github.com/catalyst/moodle-tool_objectfs/pull/300/files

PR #124, opened 2019-01-22
WR #305365: Add tweak to allow the filesystem to be set from the config

pauldamiani:


brendanheywood:
@pauldamiani what is the context behind this? It seems like an anti-feature to me

pauldamiani:
@brendanheywood This branch and PR can be deleted. It's no longer relevant

PR #139, opened 2019-06-03
Process candidates in small batches

dmitriim:


classes/object_manipulator/manipulator.php
```diff
@@ -84,58 +84,60 @@ public function __construct($filesystem, $config) {
...
         if (!$this->manipulator_can_execute()) {
             mtrace('Objectfs manipulator exiting early');
             return;
         }
 
-        if (count($objectrecords) == 0) {
-            mtrace('No candidate objects found.');
-            return;
-        }
+        $limit = 1000; // TODO: config option?
```
brendanheywood: Is there a point to this? I'd just remove this and let the default arg happen

If we add a config then I'd add into get_candidate_objects so that code can override it by setting the arg

classes/object_manipulator/manipulator.php
```diff
@@ -84,58 +84,60 @@ public function __construct($filesystem, $config) {
...
 
-        $this->logger->start_timing();
+        while (time() <= $this->finishtime) {
+            $objects = $this->get_candidate_objects($limit);
```
brendanheywood: This double nest loop looks wrong to me.

If we run out of the first batch of 1000 records then shouldn't we just exit and we'll pick up the next 1000 on the next cron tick?

classes/object_manipulator/manipulator.php
dmitriim: I wanted to make it more obvious

PR #147, opened 2019-06-20
File deletion/trashdir enhancement

golenkovm:
File deletion (moving to trashdir) and recovery includes:
- When local file is deleted it goes to local trash dir.
- When external file is deleted it doesn't pull back locally anymore. If `$CFG->tool_objectfs_delete_externally = true` then external file moves to external trash dir. if not - remains in its previous path.
- When duplicated file is deleted it removes locally. If `$CFG->tool_objectfs_delete_externally = true` then external file moves to external trash dir. If not - remains in its previous path.
- Core `recover_file` method was extended to recover external file from external trash dir. 
- PHP function `unlink()` was replaced by method and extended in client class: `unlink()` for S3 storage and `BlobRestProxy->deleteBlob` for Azure storage.
- Tests added to cover new functionality.

### DIFF
--- a/classes/client/s3_client.php
+++ b/classes/client/s3_client.php
@@ -120,9 +120,11 @@ class s3_client implements object_client {
     }
 
     public function verify_object($contenthash, $localpath) {
-        $localmd5 = md5_file($localpath);
-        $externalmd5 = $this->get_md5_from_hash($contenthash);
-        if ($externalmd5) {
+        // For objects uploaded to S3 storage using the multipart upload, the etag will not be the objects MD5.
+        // So we can't compare here to verify the object.
+        // For now we just check that we can retrieve any Etag to verify the object for all supported storages.
+        $retrievemd5 = $this->get_md5_from_hash($contenthash);
+        if ($retrievemd5) {
             return true;
         }
         return false;
@@ -137,6 +139,36 @@ class s3_client implements object_client {
     public function get_fullpath_from_hash($contenthash) {
         $filepath = $this->get_filepath_from_hash($contenthash);
         return "s3://$this->bucket/$filepath";
+    }
+
+    /**
+     * Returns s3 trash fullpath to use with php file functions.
+     *
+     * @param  string $contenthash contenthash used as key in s3.
+     * @return string trash fullpath to s3 object.
+     */
+    public function get_trash_fullpath_from_hash($contenthash) {
+        $filepath = $this->get_filepath_from_hash($contenthash);
+        return "s3://$this->bucket/trash/$filepath";
+    }
+
+    /**
+     * Deletes a file in S3 storage.
+     *
+     * @path   string full path to S3 file.
+     */
+    public function delete_file($fullpath) {
+        unlink($fullpath);
+    }
+
+    /**
+     * Moves a file in S3 storage.
+     *
+     * @param string $currentpath     current full path to S3 file.
+     * @param string $destinationpath destination path.
+     */
+    public function rename_file($currentpath, $destinationpath) {
+        rename($currentpath, $destinationpath);
     }
 
     /**
--- a/classes/object_file_system.php
+++ b/classes/object_file_system.php
@@ -28,6 +28,9 @@
 
 namespace tool_objectfs;
 
+use stored_file;
+use file_storage;
+use BlobRestProxy;
 defined('MOODLE_INTERNAL') || die();
 
 require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
@@ -150,6 +153,10 @@ abstract class object_file_system extends \file_system_filedir {
         return $this->externalclient->get_fullpath_from_hash($contenthash);
     }
 
+    protected function get_external_trash_path_from_hash($contenthash) {
+        return $this->externalclient->get_trash_fullpath_from_hash($contenthash);
+    }
+
     protected function get_external_path_from_storedfile(\stored_file $file) {
         return $this->get_external_path_from_hash($file->get_contenthash());
     }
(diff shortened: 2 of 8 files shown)


classes/client/s3_client.php
```diff
@@ -139,6 +139,17 @@ public function get_fullpath_from_hash($contenthash) {
...
 
+    /**
+     * Returns s3 trash fullpath to use with php file functions.
+     *
+     * @param  string $contenthash contenthash used as key in s3.
+     * @return string trash fullpath to s3 object.
+     */
+    public function get_trash_fullpath_from_hash($contenthash) {
+        $filepath = $this->get_filepath_from_hash($contenthash);
+        return "s3://$this->bucket/trash/$filepath";
```
dmitriim: looks like there some code duplication in get_fullpath_from_hash and  get_trash_fullpath_from_hash.  Probably the cod e can be updated to eliminate it.

classes/object_file_system.php
```diff
@@ -472,14 +489,18 @@ public function remove_file($contenthash) {
...
      */
-    public function copy_file_from_hash_to_path($contenthash, $destinationpath) {
-        $path = $this->get_remote_path_from_hash($contenthash);
-        copy($path, $destinationpath);
+    public function copy_external_file_to_trashdir_from_hash($contenthash) {
+        if ($this->deleteexternally) {
+            $currentpath = $this->get_external_path_from_hash($contenthash);
+            $destinationpath = $this->get_external_trash_path_from_hash($contenthash);
+
+            return copy($currentpath, $destinationpath);
```
dmitriim: Don't you think it's better to use existing  copy_file_from_hash_to_path method here?

classes/object_file_system.php
```diff
@@ -492,15 +513,27 @@ public function delete_object_from_hash($contenthash) {
...
             case OBJECT_LOCATION_EXTERNAL;
-                $this->delete_external_file_from_hash($contenthash);
+                $success = $this->copy_external_file_to_trashdir_from_hash($contenthash);
+                if ($success) {
+                    $this->delete_external_file_from_hash($contenthash);
+                }
                 break;
+
+            case OBJECT_LOCATION_ERROR;
+                return;
```
dmitriim: I'd recommend to add default option  as well here.

tests/object_file_system_test.php
```diff
@@ -433,7 +432,6 @@ public function test_object_storage_deleter_can_delete_object_if_enabledelete_is
         $this->delete_draft_files($filehash);
         $this->filesystem->remove_file($filehash);
         $this->assertFalse($this->is_locally_readable_by_hash($filehash));
-        $this->assertFalse($this->is_externally_readable_by_hash($filehash));
     }
 
     public function test_object_storage_deleter_can_delete_object_if_enabledelete_is_off_and_object_is_remote() {
```
kenneth-hendricks: I don't see any tests to cover the new functionality

classes/object_file_system.php
```diff
@@ -492,15 +513,27 @@ public function delete_object_from_hash($contenthash) {
...
+
             case OBJECT_LOCATION_EXTERNAL;
-                $this->delete_external_file_from_hash($contenthash);
+                $success = $this->copy_external_file_to_trashdir_from_hash($contenthash);
+                if ($success) {
+                    $this->delete_external_file_from_hash($contenthash);
+                }
                 break;
+
+            case OBJECT_LOCATION_ERROR;
```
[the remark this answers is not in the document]
golenkovm: Added in further commit

classes/client/s3_client.php
golenkovm: These two functions are different. They return different paths - path to file and path to file when it's deleted and moved to trash.

classes/object_file_system.php
golenkovm: copy_file_from_hash_to_path method invokes get_remote_path_from_hash that has a few conditions and slightly different behaviour than copy_external_file_to_trashdir_from_hash.

golenkovm: copy_file_from_hash_to_path method invokes get_remote_path_from_hash that has a few conditions and slightly different behaviour than copy_external_file_to_trashdir_from_hash.

classes/object_file_system.php
golenkovm: Added in further commit.

tests/object_file_system_test.php
golenkovm: Added in further commit.

classes/client/s3_client.php
```diff
@@ -122,7 +122,7 @@ private function get_md5_from_hash($contenthash) {
     public function verify_object($contenthash, $localpath) {
         $localmd5 = md5_file($localpath);
         $externalmd5 = $this->get_md5_from_hash($contenthash);
-        if ($externalmd5) {
+        if ($externalmd5 === $localmd5) {
```
kenneth-hendricks: So this is what we used to do, but the code was changed to just check that we could get ant kind of externalmd5 when we discovered the undocumented change that the etag that AWS stores is not always the md5. So this check essentially became an is readable check. Making this change will actually break this plugin for a subset of objects.

kenneth-hendricks: See https://forums.aws.amazon.com/thread.jspa?messageID=203436&#203436

brendanheywood: @kenneth-hendricks is this true of all stores? And was this introduced when we implemented multi part upload, or on the aws side?

@golenkovm whatever the outcome, bake this into a comment please

golenkovm: @brendanheywood verify_object method changed for all storages, the note is added

Issue #140, opened 2019-06-03
Reduce lock waiting when run cron task


dmitriim:
When we process candidates we check for the locks each time and wait for 600 seconds. Probably we don't need to wait that long.

golenkovm:
Hey @brendanheywood 
What do you think about moving it to config? Let it be a variable with default value 0.

brendanheywood:
I would never wait at all, the only reason there would ever be an existing lock is if another process is doing the same thing as what this process is wanting to do. So just set it to zero and ignore it and move on. If there is something valid that needs to be done then it will get picked up on the next cron tick. I don't think config has any real value either, this should not be something you would ever want to tune to non-zero

golenkovm:
Changes after discussion:
- Remove lock waiting when run cron task.
- Add lock timing and status logging if the plugin has logging turned on.
- Add a test to acquire a lock if object is not locked.

Merged into master.

PR #150, opened 2019-07-08
Remove lock waiting when run cron task

golenkovm:
Issue #140:
- Remove lock waiting when run cron task.
- Add lock timing and status logging if the plugin has logging turned on.
- Add a test to acquire a lock if object is not locked.

### DIFF
--- a/classes/log/real_time_logger.php
+++ b/classes/log/real_time_logger.php
@@ -90,4 +90,13 @@ class real_time_logger extends objectfs_logger {
         error_log($logstring);
         // @codingStandardsIgnoreEnd
     }
+
+    public function log_lock_timing($lock) {
+        $locktime = $this->get_timing();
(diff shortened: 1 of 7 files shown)


classes/log/real_time_logger.php
```diff
@@ -90,4 +90,13 @@ public function log_object_query($queryname, $objectcount, $objectsum = 0) {
         error_log($logstring);
         // @codingStandardsIgnoreEnd
     }
+
+    public function log_lock_timing($lock) {
```
brendanheywood: this function is duplicated, can it be shared in a base class somewhere?

golenkovm: Refactored in next commit. 
log_lock_timing method moved to objectfs_logger class and extended in null_logger class as a dummy.

PR #154, opened 2019-07-10
Swift-object-storage merge fix

golenkovm:
Updated the code to make it compatible with PHP 5.6 as it was originally pure PHP 7.


classes/client/swift_client.php
```diff
@@ -308,6 +308,7 @@ public function define_swift_check($mform, $config) {
 
         } else {
             $mform->addElement('html', $OUTPUT->notification($connection->message, 'notifyproblem'));
+            $permissions = true;
```
brendanheywood: it this line intended?

golenkovm: Following our discussion this will be fixed when issue #155 is solved.

PR #158, opened 2019-07-16
Create 'batchsize' config variable

golenkovm:


classes/form/settings_form.php
```diff
@@ -108,6 +108,10 @@ public function define_file_transfer_section($mform, $config) {
         $mform->addHelpButton('sizethreshold', 'settings:sizethreshold', 'tool_objectfs');
         $mform->setType("sizethreshold", PARAM_INT);
 
+        $mform->addElement('text', 'batchsize', get_string('settings:batchsize', 'tool_objectfs'));
+        $mform->addHelpButton('batchsize', 'settings:batchsize', 'tool_objectfs');
+        $mform->setType("batchsize", PARAM_INT);
+
```
brendanheywood: I think you are missing a setDefault() which should also be 1000?

classes/object_manipulator/manipulator.php
```diff
@@ -49,15 +49,18 @@ abstract class manipulator {
...
      *
      * @param object_file_system $filesystem object file system
      * @param int $maxruntime What time the file manipulator should finish execution by
      */
     public function __construct($filesystem, $config) {
-         $this->finishtime = time() + $config->maxtaskruntime;
-         $this->filesystem = $filesystem;
+        $this->finishtime = time() + $config->maxtaskruntime;
+        $this->filesystem = $filesystem;
+        $this->batchsize = $config->batchsize;
```
brendanheywood: Also protect this and default to 10000 if the config is not set. This is not guaranteed when you run an upgrade from the cli, it's only later that this gets set in the gui

brendanheywood: Also is there any current place where an admin can easily see a history of how long each cron tick took, including the time for the sql vs the time spent doing the transfer and how many it did?

This seems like a good thing to push into a moodle log record just a basic summary of this data

classes/form/settings_form.php
golenkovm: Default value is set in `lib.php`:
https://github.com/catalyst/moodle-tool_objectfs/pull/158/commits/cc1b80cec00bb9360a405f59b702e6db4e1785d2#diff-828043dc9f8d10ccd58ff8b96297794fR88

classes/object_manipulator/manipulator.php
golenkovm: Default value is set and protected on `lib.php`.

Regarding logging we can create an Issue and add it to TODO project. 
I'd suggest to store this data on a database layer. It will give us a chance to easily select the data for a period, build a graph, take a look at data dynamic, etc.

classes/form/settings_form.php
brendanheywood: I still think the formslib should be told what the default is, even if the default will be set before being loaded into the form. The default value gets shown as a hint in the form UI this way.

golenkovm: As discussed we leave it as is.

Issue #45, opened 2017-02-22
Investigate the viability of overriding xsendfile for S3 to redirect to signed S3 url


brendanheywood:
* [ ] This means every file.php or pluginfile.php request will serve a redirect to S3 so it will have a small round trip penalty, but the files will stream directly to the browser instead of being proxied at the php level.

https://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-presigned-url.html

* [ ] Also investigate the use of ```$CFG->urlrewriteclass``` as a means of rewriting all urls which reference these urls so we can avoid the redirect. We must only conditionally rewrite urls which we know have been pushed to the object store.

* [ ] If you implement a url rewriter it would ne nice to pass in a fallback urlrewritter so that we can maintain both clean urls and S3 rewrites side by side

* [ ] think about the default expiry time limit for files. One could argue for very short like 1 minute, or something much longer like re-use the session max age setting in moodle which is usually 2-4 hours

brendanheywood:
A few quick thoughts from a very brief scan of the aws signing docs:

1) we will need to increase the timeout, the default is 15 mins, I think this should be roughly the same order of magnitude as the moodle session. In particular range requests for media files which are longer than 15 minutes will fail.

2) For the use case where a signature fails, perhaps instead of S3 serving a 403 it could just deterministically redirect back to the original moodle url. I have no idea is S3 even supports this so it may not be viable. If that works then it would bounce to moodle, and then get re-signed and bounce back to S3 again.

brendanheywood:
It looks like we can only abuse the xsendfile if xsendfile is on everywhere which includes files outside the file api, so turning this on in s3 would break local files. So this is a no go without another core change even if we did want this

PR #165, opened 2019-07-31
Overriding xsendfile method to redirect to pre-signed S3 URL

brendanheywood:
Issue #45 

What has been done:
- Implement pre-singing methos for S3 storage.
- Implement redirect to pre-signed url if file stored externally.
- Change file name from contenthash to human-readable (original) name.
- Send disposition and content header only if Moodle tried to set it.
- Testing page for Pre-Signed URLs.


classes/client/object_client.php
```diff
@@ -27,15 +27,34 @@
...
+
+    /**
+     * Generates pre-signed URL to storage file from its hash.
+     *
+     * @param string $contenthash File content hash.
+     *
+     * @return string.
+     */
+    public function generate_signed_url($contenthash) {
+        return 'Not supported';
```
brendanheywood: should we just throw an exception here?

classes/object_file_system.php
```diff
@@ -661,4 +661,23 @@ protected function copy_from_external_to_local($contenthash) {
...
+     *
+     * @return string.
+     */
+
+    public function generate_signed_url_to_external_file_from_hash($contenthash) {
+
+        if ($this->externalclient->support_signed_urls()) {
+            $signedurl = $this->externalclient->generate_signed_url($contenthash);
+        } else {
+            $signedurl = 'Pre-Signed URLs not supported for current storage';
```
brendanheywood: I think remove this and return null, we never want that string accidentally leaking in a url rewrite

classes/object_file_system.php
```diff
@@ -661,4 +663,25 @@ protected function copy_from_external_to_local($contenthash) {
...
+        return $this->externalclient->generate_signed_url($contenthash, $filename);
+    }
+
+    /**
+     * Returns file name from globals variable.
+     *
+     * @return string.
+     */
+    public function get_filename_from_globals() {
+        $pos = strrpos($GLOBALS['ME'], '/');
```
brendanheywood: perhaps basename?

https://www.php.net/manual/en/function.basename.php

classes/client/object_client.php
```diff
@@ -42,19 +42,57 @@ public function register_stream_wrapper() {
...
+    /**
+     * Checks that Content-Disposition represented in headers.
+     *
+     * @param array $headers request headers.
+     *
+     * @return bool
+     */
+    public function content_disposition_exists_in_headers($headers) {
+        foreach ($headers as $header) {
+            $found = strpos($header, 'Content-Disposition');
```
brendanheywood: It's more complicated than this, it's not a boolean header.

```
Content-Disposition: attachment; filename="cool.html"
Content-Disposition: inline

```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

We need to actually parse the contents of the header, pull it apart, and then reuse parts of it in difference ways:

1) the filename="filename.jpg" page, we could use this instead of parsing the request url. Really we should do both, and this one takes precendence

2) the 'attachment' vs 'inline', we should pass this along to aws to. Ideally if we can pass the entire header over the AWS and get them to add it? no idea if this is possible

classes/client/object_client.php
```diff
@@ -42,19 +42,57 @@ public function register_stream_wrapper() {
...
+
+    /**
+     * Returns Content-Disposition header from headers.
+     *
+     * @param array $headers request headers.
+     *
+     * @return string file name.
+     * @throws \coding_exception
+     */
+    public function get_content_disposition_header($headers) {
```
brendanheywood: Is there much value in having the two functions?
content_disposition_exists_in_headers vs get_content_disposition_header

It could just return an empty string if there are none, and it's a falsy value

brendanheywood:
Next steps:

* [ ] 0) make a new admin testing page which tests all of the below, but add the tests 1 by one as each feature is implemented. Some will be links on the page, some might be an embedded image or iframe etc it will be case by case and grow over time. Some of this is hard to unit tests so I want it to be something we can run in prod or test at any point.

The first time this page is loaded, it should read a static file from the plugin code and push it into the file api. We will need a few test files, an image, a pdf, a video file, etc as needed. We will need to implement a pluginfile callback for these files which handle all the testing options

* [ ] 1) add more nuanced settings so we only redirect if the file is > a certain configurable size. Should default to 0 which means never redirect. Add test files for a tiny embedded image file, and a large pdf link + video link

* [ ] 2) when redirecting, read the caching expires headers that moodle has sent, and still send those when we redirect. We might get this for free but test it

* [ ] 3) if the headers have cache: public or cache immutable etc, then we want to sign the url with no expiry ever, and also send those same cache / expiry headers for the signed url too. We need to test to ensure that these signed urls themselves don't change, ie does the signature have a timestamp mixed into it?

* [ ] 4) if the headers have an expiry of any value eg 10 mins or 1 hour, then set the expiry of the signature to be the length of the $CFG->sessiontimeout . We need this to be long to support using the back button and then forward again. Better yet turn this in into it's own config, ie a signatureexpiry duration

PR #175, opened 2019-08-26
Cron tasks optimization

golenkovm:
On big sites, cron tasks take a long time to complete and uses a lot of resources todo so.
This is because they use _left join_ and _is not NULL_ condition. 
Same query, but with inner join and without _is not NULL_ condition works much faster.

What should be implemented:
- [ ] Add a record to _tool_objectfs_objects_ table when new file created;
- [ ] Amend SQLs to use _inner join_ and remove _o.location is NULL_ condition;
- [ ] Add new cron task that would check files that do not exist in _tool_objectfs_objects_ and add them with current location.

### DIFF
--- a/classes/local/object_manipulator/checker.php
+++ b/classes/local/object_manipulator/checker.php
@@ -51,7 +51,7 @@ class checker extends manipulator {
 
     protected function get_candidates_sql() {
         $sql = 'SELECT f.contenthash
-                  FROM {files} f 
+                  FROM {files} f
              LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
                  WHERE f.filesize > 0
                    AND o.location is NULL
--- a/classes/local/object_manipulator/checker.php
+++ b/classes/local/object_manipulator/checker.php
@@ -27,12 +27,11 @@ namespace tool_objectfs\local\object_manipulator;
 
 defined('MOODLE_INTERNAL') || die();
 
-require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
-
 class checker extends manipulator {
 
     /**
-     * Pusher constructor.
+     * Checker constructor.
+     * This manipulator adds location for files that do not have records in {tool_objectfs_objects} table.
      *
      * @param object_client $client remote object client
      * @param object_file_system $filesystem objectfs file system
@@ -52,9 +51,10 @@ class checker extends manipulator {
 
     protected function get_candidates_sql() {
         $sql = 'SELECT f.contenthash
-                  FROM mdl_files f LEFT JOIN mdl_tool_objectfs_objects o ON f.contenthash = o.contenthash
+                  FROM {files} f 
+             LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
                  WHERE f.filesize > 0
-                       AND o.location is NULL
+                   AND o.location is NULL
               GROUP BY f.contenthash';
 
         return $sql;
@@ -62,7 +62,6 @@ class checker extends manipulator {
 
     protected function get_candidates_sql_params() {
         $params = array();
-
         return $params;
     }
 
--- a/classes/local/object_manipulator/checker.php
+++ b/classes/local/object_manipulator/checker.php
@@ -1,0 +1,73 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Checks files to have their location stored in database.
+ *
+ * @package   tool_objectfs
+ * @author    Mikhail Golenkov <mikhailgolenkov@catalyst-au.net>
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace tool_objectfs\local\object_manipulator;
+
+defined('MOODLE_INTERNAL') || die();
+
+require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
+
+class checker extends manipulator {
+
+    /**
+     * Pusher constructor.
+     *
+     * @param object_client $client remote object client
+     * @param object_file_system $filesystem objectfs file system
+     * @param object $config objectfs config.
+     */
+    public function __construct($filesystem, $config, $logger) {
+        parent::__construct($filesystem, $config);
+
+        $this->logger = $logger;
+        // Inject our logger into the filesystem.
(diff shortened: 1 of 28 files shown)


classes/local/object_manipulator/checker.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->filesystem->set_logger($this->logger);
+    }
+
+    protected function get_query_name() {
+        return 'get_check_candidates';
+    }
+
+    protected function get_candidates_sql() {
+        $sql = 'SELECT f.contenthash
+                  FROM mdl_files f LEFT JOIN mdl_tool_objectfs_objects o ON f.contenthash = o.contenthash
```
brendanheywood: mdl_files vs {files}

classes/local/object_manipulator/checker.php
```diff
@@ -0,0 +1,73 @@
...
+                 WHERE f.filesize > 0
+                       AND o.location is NULL
```
brendanheywood: weird whitespace?

classes/local/object_manipulator/checker.php
```diff
@@ -0,0 +1,73 @@
...
+namespace tool_objectfs\local\object_manipulator;
+
+defined('MOODLE_INTERNAL') || die();
+
+require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
+
+class checker extends manipulator {
+
+    /**
+     * Pusher constructor.
```
brendanheywood: php docs looks like cut and paste, would be good to explain the purpose of the checker here

classes/task/check_objects_location.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace tool_objectfs\task;
+
+use tool_objectfs\local\object_manipulator\manipulator;
+
+defined('MOODLE_INTERNAL') || die();
+
+require_once( __DIR__ . '/../../lib.php');
+require_once(__DIR__ . '/../../../../../config.php');
```
brendanheywood: this looks wrong?

db/tasks.php
```diff
@@ -71,5 +71,14 @@
         'dayofweek' => '*',
         'month'     => '*'
     ),
+    array(
+        'classname' => 'tool_objectfs\task\check_objects_location',
+        'blocking'  => 0,
+        'minute'    => '7',
```
brendanheywood: 7 seems a little arbitrary, what is the purpose of this? If it is to spread load and not hit the 0th minute then a value of "R" for random is probably better

https://docs.moodle.org/dev/Task_API#Cron_syntax_examples

classes/local/object_manipulator/checker.php
brendanheywood: Also the LEFT JOIN should be on a new line

classes/local/object_manipulator/checker.php
```diff
@@ -0,0 +1,73 @@
...
+ * @author    Mikhail Golenkov <mikhailgolenkov@catalyst-au.net>
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace tool_objectfs\local\object_manipulator;
+
+defined('MOODLE_INTERNAL') || die();
+
+require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
```
brendanheywood: is this actually needed here?

classes/local/object_manipulator/puller.php
```diff
@@ -63,12 +63,12 @@ protected function get_candidates_sql() {
                        f.contenthash,
                        MAX(f.filesize) AS filesize
                   FROM {files} f
-             LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+            INNER JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+                 WHERE f.filesize <= ?
+                       AND o.location = ?
```
brendanheywood: whitespace not aligned

classes/local/object_manipulator/deleter.php
```diff
@@ -81,13 +81,13 @@ protected function get_candidates_sql() {
                        f.contenthash,
                        MAX(f.filesize) AS filesize
                   FROM {files} f
-             LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+            INNER JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
                  WHERE o.timeduplicated <= ?
                        AND o.location = ?
+                       AND f.filesize > ?
```
brendanheywood: more whitespace, and the unchanged line above it

brendanheywood: re the whitespace indent, this is slightly confusing as the wiki page example is kinda ambiguous or misleading.

https://docs.moodle.org/dev/SQL_coding_style#Indentation

In my book the AND should be left of the red line, like this:

https://github.com/moodle/moodle/blob/master/admin/tool/cohortroles/classes/api.php#L210

Core is all over the shop so it's easy to find examples both ways. HQ need a better policy and to update the wiki

brendanheywood: fyi https://tracker.moodle.org/browse/MDL-49238?focusedCommentId=683604&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-683604

classes/local/object_manipulator/deleter.php
```diff
@@ -81,13 +81,13 @@ protected function get_candidates_sql() {
                        f.contenthash,
                        MAX(f.filesize) AS filesize
                   FROM {files} f
-             LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+            INNER JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
```
brendanheywood: is INNER redundant?

classes/local/object_manipulator/puller.php
```diff
@@ -63,12 +63,12 @@ protected function get_candidates_sql() {
                        f.contenthash,
                        MAX(f.filesize) AS filesize
                   FROM {files} f
-             LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+            INNER JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
```
brendanheywood: again is INNER needed?

lang/en/tool_objectfs.php
```diff
@@ -30,6 +30,7 @@
 $string['delete_local_objects_task'] = 'Object file system delete local objects task';
 $string['pull_objects_from_storage_task'] = 'Object file system download objects task';
 $string['recover_error_objects_task'] = 'Object error recovery task';
+$string['check_objects_location_task'] = 'Object file system check objects location task';
```
brendanheywood: I'd just explain this a little more and make it clear that it's only really important for the initial migration.

After that it can be run less frequently like once a night if performance is an issue.I'd also put this into the README in a new section called 'performance tuning'

tests/checker_test.php
```diff
@@ -0,0 +1,169 @@
...
+        $this->assertEquals(OBJECT_LOCATION_DUPLICATED, $dblocation);
+        $this->assertTrue($dblocationtype);
+    }
+
+    public function test_checker_get_location_external_if_object_is_external() {
+        global $DB;
+        $file = $this->create_remote_object();
+        $fslocation = $this->filesystem->get_object_location_from_hash($file->contenthash);
+        $dblocation = $DB->get_field('tool_objectfs_objects', 'location', array('contenthash' => $file->contenthash));
+        $dblocationtype = gettype($dblocation) == 'string' ? true : false;
```
brendanheywood: general best practice is to avoid logic in tests. Is there two different code paths and we should be exercising both here?

Issue #176, opened 2019-08-29
Implement tests for *presigned_url* methods


golenkovm:
https://github.com/catalyst/moodle-tool_objectfs/commit/a3b600eadd07ffe7645385dbbe5d3145cd930a5c

Before this commit will be merged into master branch new methods should be covered by tests:
- [ ] _presigned_url_configured()_
- [ ] _presigned_url_should_redirect()_

dmitriim:
Fixed as part of https://github.com/catalyst/moodle-tool_objectfs/pull/183

PR #177, opened 2019-08-30
Implement tests for *presigned_url* methods.

golenkovm:
Issue #176

### DIFF
--- a/tests/classes/test_client.php
+++ b/tests/classes/test_client.php
@@ -108,6 +108,7 @@ class test_client extends object_client_base {
     public function support_signed_urls() {
         return false;
     }
+
     public function generate_signed_url($contenthash) {
         return 'Not supported';
     }
--- a/tests/classes/test_client.php
+++ b/tests/classes/test_client.php
@@ -104,13 +104,5 @@ class test_client extends object_client_base {
     public function get_maximum_upload_size() {
         return 5000000000;
     }
-
-    public function support_signed_urls() {
-        return false;
-    }
-
-    public function generate_signed_url($contenthash) {
-        return 'Not supported';
-    }
 }
 
--- a/tests/object_file_system_test.php
+++ b/tests/object_file_system_test.php
@@ -537,9 +537,9 @@ class object_file_system_testcase extends tool_objectfs_testcase {
 
     public function test_presigned_url_configured_method() {
         $this->filesystem = new test_file_system();
-        $client = get_class($this->filesystem->get_external_client());
+        $externalclient = $this->filesystem->get_external_client();
 
-        if ($this->integration_client_supports_presigned_urls($client)) {
+        if ($externalclient->support_presigned_urls()) {
             $this->set_externalclient_config('enablepresignedurls', '1');
             $res = $this->filesystem->presigned_url_configured();
             $this->assertTrue($res);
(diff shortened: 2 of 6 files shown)


tests/tool_objectfs_testcase.php
```diff
@@ -252,4 +252,13 @@ protected function is_externally_readable_by_url($url) {
             return false;
         }
     }
+
+    protected function integration_client_supports_presigned_urls($client) {
+        $supportedlist = array();
+        $supportedlist[] = 'tool_objectfs\tests\test_s3_integration_client';
```
dmitriim: Can you please explain why you check that the client supports presigned urls amnually and not using support_presigned_urls method?

tests/classes/test_client.php
```diff
@@ -108,6 +108,7 @@ public function get_maximum_upload_size() {
     public function support_signed_urls() {
         return false;
     }
+
     public function generate_signed_url($contenthash) {
         return 'Not supported';
```
brendanheywood: Not quite related to your patch but we should change this to throw a 'not implemented' exception ?

golenkovm: Oh, thank you for pointing on that. We use `generate_presigned_url()` method and it throws exception. `generate_signed_url()` was the first version of this method and not used anymore. I will remove it in further commit.

tests/tool_objectfs_testcase.php
golenkovm: You're right, it's better to use `$externalclient->support_presigned_urls()` method. Re-written tests will be in further commit.

tests/object_file_system_test.php
```diff
@@ -563,35 +562,41 @@ public function test_presigned_url_should_redirect_method_return_true_regardless
...
-            $object = $this->create_local_object();
-            $this->assertFalse($this->filesystem->presigned_url_should_redirect($object->contenthash));
+        if (!$externalclient->support_presigned_urls()) {
+            $this->markTestSkipped('Pre-signed URLs not supported for given storage.');
         }
+
+        $this->set_externalclient_config('enablepresignedurls', '1');
+        $this->set_externalclient_config('presignedminfilesize', 1000);
+        $object = $this->create_local_object();
+        $this->assertFalse($this->filesystem->presigned_url_should_redirect($object->contenthash));
```
dmitriim: why is it false here? does create_local_object create the file less than 1000? If it is, can you please add some comment here?

tests/object_file_system_test.php
```diff
@@ -563,35 +562,41 @@ public function test_presigned_url_should_redirect_method_return_true_regardless
...
-            $this->set_externalclient_config('enablepresignedurls', '1');
-            $this->set_externalclient_config('presignedminfilesize', 0);
-            $object = $this->create_local_object();
-            $this->assertTrue($this->filesystem->presigned_url_should_redirect($object->contenthash));
+        if (!$externalclient->support_presigned_urls()) {
+            $this->markTestSkipped('Pre-signed URLs not supported for given storage.');
         }
+
+        $this->set_externalclient_config('enablepresignedurls', '1');
+        $this->set_externalclient_config('presignedminfilesize', 0);
```
dmitriim: Can you please also test scenarios when presignedminfilesize and enablepresignedurls are set to null, empty string, '0' as a string?

tests/object_file_system_test.php
golenkovm: It is false because local test files are created with file size = 10:
https://github.com/catalyst/moodle-tool_objectfs/blob/issue176/tests/tool_objectfs_testcase.php#L86

Issue #180, opened 2019-09-04
Image preview cause pulling images from external storages


golenkovm:
Objectfs pulls files locally each time when `get_local_path_from_hash()` method gets invoked. 
This method is used to preview images and get their width, height and mimetype. 

So even if pre-signed urls feature is enabled we have this scenario:
- user opens a page with image (preview, like in private files page);
- Objectfs pulls down the image to let Moodle check its attributes;
- Objectfs updates db and marks this file as diplicated;
- Objectfs generates pre-signed url and feeds it to user;
- then `delete` task picks this file up and deletes it locally.

If some image is frequently asked it might cause excess traffic and unnessesary work to download, update db, delete, update db and agian.

It would be good to extend stored_files `get_imageinfo()` method and replace `$path` to external url to file. `getimagesize` works fine with urls as well.

https://tracker.moodle.org/browse/MDL-70368

brendanheywood:
I just did a quick experiment, jpeg headers are at the start of the file, you can do a brutal partial byte dumb of a file and the exif info still works:

```
$ head -c 1000 babyyoda.jpeg > babyyoda.header.jpeg
$ ls -lh babyyoda.*
-rw-r--r-- 1 brendan brendan 1000 Dec 14 15:46 babyyoda.header.jpeg
-rw-r--r-- 1 brendan brendan 189K Dec 14 15:35 babyyoda.jpeg
$ exiftool babyyoda.header.jpeg 
ExifTool Version Number         : 10.80
File Name                       : babyyoda.header.jpeg
Directory                       : .
File Size                       : 1000 bytes
File Modification Date/Time     : 2020:12:14 15:46:44+11:00
File Access Date/Time           : 2020:12:14 15:43:44+11:00
File Inode Change Date/Time     : 2020:12:14 15:46:44+11:00
File Permissions                : rw-r--r--
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
Image Width                     : 1200
Image Height                    : 900
Encoding Process                : Baseline DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 3
Y Cb Cr Sub Sampling            : YCbCr4:2:0 (2 2)
Image Size                      : 1200x900
```

We should be able to just grab the stream and download 1 or 2kb into temp and run get_imageinfo on that

brendanheywood:
Just tested this with a png as well as also works with a 1k header chunk

[...]

PR #181, opened 2019-09-04
Get image info by remote path if file is external

golenkovm:
This PR covers Issue #180 .

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -753,4 +753,28 @@ abstract class object_file_system extends \file_system_filedir {
 
         return $result;
     }
+
+    /**
+     * Returns information about image.
+     * Information is determined from the file content.
+     * If file is external get image info by remote path.
+     *
+     * @param stored_file $file The file to inspect
+     * @return mixed array with width, height and mimetype; false if not an image
+     */
+    public function get_imageinfo(stored_file $file) {
+        if (!$this->is_image_from_storedfile($file)) {
+            return false;
+        }
+
+        $contenthash = $file->get_contenthash();
+        $location = $this->get_object_location_from_hash($contenthash);
+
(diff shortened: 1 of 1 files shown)


classes/local/store/object_file_system.php
```diff
@@ -753,4 +753,28 @@ public function add_file_from_string($content) {
...
+     */
+    public function get_imageinfo(stored_file $file) {
+        if (!$this->is_image_from_storedfile($file)) {
+            return false;
+        }
+
+        $contenthash = $file->get_contenthash();
+        $location = $this->get_object_location_from_hash($contenthash);
+
+        if ($location === OBJECT_LOCATION_EXTERNAL) {
```
dmitriim: I'm a bit worried about '===' here. If location will be returned as string '2' this if will return false.

classes/local/store/object_file_system.php
```diff
@@ -753,4 +753,28 @@ public function add_file_from_string($content) {
...
+
+    /**
+     * Returns information about image.
+     * Information is determined from the file content.
+     * If file is external get image info by remote path.
+     *
+     * @param stored_file $file The file to inspect
+     * @return mixed array with width, height and mimetype; false if not an image
+     */
+    public function get_imageinfo(stored_file $file) {
```
dmitriim: Would be good to see some tests to cover this function

classes/local/store/object_file_system.php
dmitriim: Actually scratch that. I looked at the code and get_object_location_from_hash will never return a string

tuanngocnguyen:
Hi @golenkovm ,

Since this has been inactive for some time, I am closing this PR.
Please create new MR against MOODLE_33_STABLE/MOODLE_310_STABLE if you are still working on the issue.
Thanks

PR #197, opened 2019-11-30
Digital ocean

kisonay:
I just upgraded my Moodle instance to the latest master branch code. I use the DigitalOcean integration and upon installing I noticed that the plugin was no longer working. Did some debugging and found that since everything was restructured, the code was looking for files in a store/do directory rather than the new store/digitalocean directory.

This pull request fixed that reference and updates the readme to reflect the new digitalocean directory. While I was making changes, I also included the new Frankfrut region as an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ $CFG->alternative_file_system_class = '\tool_objectfs\azure_file_system';
 
 * DigitalOcean Spaces
 ```php
-$CFG->alternative_file_system_class = '\tool_objectfs\do_file_system';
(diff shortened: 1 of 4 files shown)


brendanheywood:
The restructure was intended to keep backwards compatibility, so should we just rename the class back to _do_ instead of digital_ocean? I prefer the longer one, but the support seems better to me. I'm on the fence

kisonay:
I too like digitalocean vs do as it is more descriptive. I’m fine with it either way.

kisonay:
Let me know how you would like to proceed.

brendanheywood:
All good :)

PR #200, opened 2019-12-04
Master 303089 review

nickmccutcheon:


classes/local/form/settings_form.php
```diff
@@ -58,6 +58,7 @@ public function definition() {
         } else {
             // TODO: set enablepresignedurls = 0.
             $mform->setDefault("enablepresignedurls", 0);
+            $mform->setDefault("enablepresignedcloudfronturls", 0);
```
brendanheywood: I'm expecting to not see any changes to files outside of the s3 store

presignedcloudfronturl_tests.php
```diff
@@ -0,0 +1,78 @@
+<?php
```
brendanheywood: I don't think this file needs to exist. The testing of signed files should be generic regardless of how they are signed so the existing page should be sufficient

classes/local/form/settings_form.php
```diff
@@ -171,20 +172,20 @@ public function define_presignedurl_section($mform) {
         global $OUTPUT;
 
         $mform->addElement('header', 'presignedurlheader',
-            get_string('settings:presignedurl:header', 'tool_objectfs'));
+            get_string('settings:presignedurls3:header', 'tool_objectfs'));
```
brendanheywood: please revert this

classes/local/form/settings_form.php
```diff
@@ -171,20 +172,20 @@ public function define_presignedurl_section($mform) {
...
         $mform->setExpanded('presignedurlheader');
 
         $link = \html_writer::link(new \moodle_url('/admin/tool/objectfs/presignedurl_tests.php'),
             get_string('presignedurl_testing:page', 'tool_objectfs'));
 
         $text = $OUTPUT->heading(get_string('settings:presignedurl:warning', 'tool_objectfs').$link, 6);
         $mform->addElement('html', $OUTPUT->notification($text, 'warning'));
-
+/*
```
brendanheywood: why is this commented?

classes/local/store/object_file_system.php
```diff
@@ -677,8 +677,9 @@ public function redirect_to_presigned_url($contenthash) {
 
     public function presigned_url_configured() {
         return $this->externalclient->support_presigned_urls()
-            && $this->externalclient->enablepresignedurls
-            && isset($this->externalclient->presignedminfilesize);
+            && ($this->externalclient->enablepresignedurls
```
brendanheywood: this should not be needed, it should be encapsulated into the 3s store

classes/local/store/object_file_system.php
```diff
@@ -711,7 +712,7 @@ public function presigned_url_should_redirect($contenthash) {
      * @return string.
      */
     public function generate_presigned_url_to_external_file($contenthash, $headers = array()) {
-        return $this->externalclient->generate_presigned_url($contenthash, $headers);
+        return ($this->externalclient->generate_presigned_url($contenthash, $headers));
```
brendanheywood: please revert this

classes/local/store/s3/client.php
```diff
@@ -32,8 +32,10 @@
 use Aws\S3\ObjectUploader;
 use Aws\S3\S3Client;
 use Aws\S3\Exception\S3Exception;
+use Aws\CloudFront\CloudFrontClient;
 use tool_objectfs\local\store\object_client_base;
 
+
```
brendanheywood: remove whitespace change

classes/local/form/settings_form.php
```diff
@@ -180,11 +180,6 @@ public function define_presignedurl_section($mform) {
         $text = $OUTPUT->heading(get_string('settings:presignedurl:warning', 'tool_objectfs').$link, 6);
         $mform->addElement('html', $OUTPUT->notification($text, 'warning'));
 
-        $mform->addElement('advcheckbox', 'enablepresignedurls',
```
brendanheywood: this should not have changed, it's the master on off switch regardless of which store is used

lang/en/tool_objectfs.php
```diff
@@ -141,8 +144,21 @@
...
+$string['settings:presignedcloudfronturl:enablepresignedcloudfronturls_help'] = 'Enable CLOUDFRONT Pre-Signed URLs by setting up a Cloudfront Distribution profile at AWS.';
+$string['settings:presignedcloudfronturl:cloudfront_resource_domain'] = 'DOMAIN (inc. https://)';
+$string['settings:presignedcloudfronturl:cloudfront_resource_domain_help'] = 'Enter the domain name from which resources are requested at Cloudfront (refer to AWS Cloudfront Distribution)';
+$string['settings:presignedcloudfronturl:cloudfront_key_pair_id'] = 'Key_Pair ID from AWS';
+$string['settings:presignedcloudfronturl:cloudfront_key_pair_id_help'] = 'This is generated using AWS account \'root\' user (along with the private key .pem file).';
+$string['settings:presignedcloudfronturl:cloudfront_private_key_pem_file_pathname'] = 'PRIVATE Key .pem pathname';
+$string['settings:presignedcloudfronturl:cloudfront_private_key_pem_file_pathname_help'] = 'Full path and filename to the .pem file. Example: ' . $CFG->dataroot . '/ (CFG->dataroot) should be prepended to .pem filename. The actual .pem file must be available for file_get_content().';
```
brendanheywood: So you couldn't have known this, but this will cause a translation issue. Moodle has a way to pass in vars to lang strings eg:

$string['accesskey'] = 'Access key, {$a}';

https://docs.moodle.org/dev/String_API#Run-time_parameters

Also I would strong suggest to NOT prepend the dataroute. In fact I would go one step further and say that this must be at a fixed path eg:
/sitedata/objectfs/cloudfront.pem

Then there is one less setting and one less thing to go wrong.

classes/local/store/s3/client.php
```diff
@@ -430,10 +500,104 @@ public function generate_presigned_url($contenthash, $headers) {
...
+                    '&response-content-type=' . rawurlencode($originalcontenttype);
+
+                // Alternative without filename: $newkey = $key . '?response-content-disposition='.rawurlencode($contentdisposition.';'.(utf8_encode($originalfilename)).'').'&response-content-type='.rawurlencode($originalcontenttype);.
+
+                $resourcekey = $resourcedomain . '/' . $newkey;
+            }
+        }
+
+        if (isset($cdnconfig->expirationtime) && !empty($cdnconfig->expirationtime)) {
+            $expirationvalue = time() + $cdnconfig->expirationtime;
```
brendanheywood: this feels wrong, on line 585 we will end up with:

 time() + time() + $cdnconfig->expirationtime

brendanheywood: I'd ditch $expirationvalue and just move $expires up in the if else

lib.php
```diff
@@ -92,6 +92,8 @@ function get_objectfs_config() {
     $config->expirationtime = 10 * MINSECS;
     $config->presignedminfilesize = 0;
 
+    $config->presignurlschoices = array(0 => 'Disabled', 1 => 'S3', 2 => 'Cloudfront');
```
brendanheywood: avoid these hard coded lang strings too

brendanheywood: also this doesn't seem to be used anywhere else in the patch?

lib.php
```diff
@@ -178,3 +189,8 @@ function tool_objectfs_cron() {
 
     return true;
 }
+
+function tool_objectfs_get_signingmethod_list() {
+    $availablemethods = array(''=>'S3', 'CF'=>'Cloudfront');
```
brendanheywood: I'd expect the two labels to be lang strings here

classes/local/store/s3/client.php
```diff
@@ -430,10 +453,103 @@ public function generate_presigned_url($contenthash, $headers) {
...
+        $resourcekey = $resourcedomain . '/' . $key;
+
+        if ($nicefilename) {
+            // We are trying to deliver original filename rather than hash filename to client.
+
+            $contentdisposition = '';
+            $originalfilename = '';
+            $originalcontenttype = '';
+
+            $contentdisposition = $this->get_header($headers, 'Content-Disposition');
```
brendanheywood: I think these whole 4-5 lines can be much simpler as just:

$contentdisposition = trim($this->get_header($headers, 'Content-Disposition'));

I think the if is redundant. Same might apply to contenttype

PR #201, opened 2019-12-06
Fixing swift integration and adding swift unit test

ledship:
Fixes the swift integration and adds a unit test for swift.


brendanheywood:
@ledship as I can't test this I'm happy for you to merge at will anything purely related to just the swift store and go through whatever review process you want in nz

PR #204, opened 2020-01-01
Add new task to delete empty folders from the filedir folder

gleimermora:


classes/task/delete_local_empty_directories.php
```diff
@@ -0,0 +1,59 @@
...
+    public function execute() {
+        global $CFG;
+        $config = get_objectfs_config();
+
+        if (!tool_objectfs_should_tasks_run()) {
+            mtrace(get_string('not_enabled', 'tool_objectfs'));
+            return;
+        }
+        $filesystem = new $config->filesystem();
+        $filedirectory = $CFG->dataroot . '/filedir';
```
brendanheywood: please use $CFG->filedir

db/tasks.php
```diff
@@ -53,6 +53,15 @@
         'dayofweek' => '*',
         'month'     => '*'
     ),
+    array(
+        'classname' => 'tool_objectfs\task\delete_local_empty_directories',
+        'blocking'  => 0,
+        'minute'    => '*',
```
brendanheywood: I suspect this task will take probably 10 minutes to run, I'd lean more towards running this once a day

classes/task/delete_local_empty_directories.php
```diff
@@ -0,0 +1,59 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Task that pushes files to S3.
```
brendanheywood: cut and paste stuff

.travis.yml
```diff
@@ -39,11 +40,47 @@ env:
   - DB=mysqli MOODLE_BRANCH=MOODLE_35_STABLE
   - DB=pgsql MOODLE_BRANCH=MOODLE_36_STABLE
   - DB=mysqli MOODLE_BRANCH=MOODLE_36_STABLE
+  - DB=pgsql MOODLE_BRANCH=MOODLE_37_STABLE
```
brendanheywood: If you add more combinations to travis it can slow it right down. Your test run was:

 Ran for 29 min 51 sec
 Total time 1 hr 55 min 20 sec

There was 52 builds in the matrix. We want to make sure we have covered every db, php version and stable, but we don't need to test every combination of all of the above. We want to get the build cycle back to around 20 mins and keep it there or less.

So can you please weed out some of the combinations which are less interesting, perhaps keep all the LTS versions ie 2.7, 3.1 and 3.5 and then everything after 3.5

brendanheywood: Feel free to do that in a follow up issue too

classes/local/store/object_file_system.php
```diff
@@ -316,6 +321,50 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
         return $finallocation;
     }
 
+    /**
```
brendanheywood: missing php doc which explains what is going on here

classes/local/store/object_file_system.php
```diff
@@ -316,6 +321,50 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
...
+     * @param string $rootpath
+     */
+    public function delete_empty_folders($rootpath) {
+        if (!is_dir($rootpath)) {
+            return;
+        }
+        if ($this->is_dir_empty($rootpath)) {
```
brendanheywood: This feels like is should be below the iterator below. ie delete all my empty children dirs first, and then I can delete myself

classes/local/store/object_file_system.php
```diff
@@ -316,6 +321,50 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
...
+            rmdir($rootpath);
+            return;
+        }
+        $iterator = new RecursiveDirectoryIterator($rootpath);
```
brendanheywood: It's not clear from the docs if RecursiveDirectoryIterator is top down or bottom up. It must really be bottom up, and if it is then it should in theory delete the root itself last without special handling.

I'd like to see unit tests around this

classes/local/store/object_file_system.php
```diff
@@ -316,6 +321,50 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
...
+                rmdir($iterator->getRealPath());
+            } else if ($this->is_dir_empty($dirpath)) {
+                // We found an empty directory.
+                rmdir($dirpath);
+            }
+        }
+    }
+
+    /**
+     * @param string $foldername
```
brendanheywood: Please add the php doc, eg is folder name a full path or relative?

classes/local/store/object_file_system.php
```diff
@@ -450,7 +499,6 @@ public function remove_file($contenthash) {
             // Don't remove the file - it's still in use.
             return;
         }
-
```
brendanheywood: typo?

classes/task/delete_local_empty_directories.php
```diff
@@ -0,0 +1,61 @@
...
+
+        if (!tool_objectfs_should_tasks_run()) {
+            mtrace(get_string('not_enabled', 'tool_objectfs'));
+            return;
+        }
+        $filesystem = new $config->filesystem();
+        if (!isset($CFG->filedir)) {
+            $CFG->filedir = $CFG->dataroot . '/filedir';
+        }
+        $filesystem->delete_empty_folders($CFG->filedir);
```
brendanheywood: We will need to get some stats on how slow this is on a big filedir

classes/local/store/object_file_system.php
```diff
@@ -298,11 +302,12 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
 
         if ($initiallocation === OBJECT_LOCATION_DUPLICATED) {
             $localpath = $this->get_local_path_from_hash($contenthash);
-
             if ($this->verify_external_object_from_hash($contenthash)) {
                 $success = unlink($localpath);
 
                 if ($success) {
+                    // Remove dir if empty.
+                    $this->delete_empty_folders(dirname($localpath));
```
brendanheywood: This only seems half right. Lets say we delete the last file eg

filedir/aa/bb/aabbccddeeff

now we are left with

fileedir/aa/bb/

then delete_empty_folders will delete the parent:

fileedir/aa/

but what about the grandparent? If it is empty we want to delete it too, but we should only attempt that if we deleted the first one, ie iterate up the ancestor tree. We should be careful to not iterate outside of the filedir though

.travis.yml
gleimermora: updated

classes/local/store/object_file_system.php
gleimermora: doc updated

classes/local/store/object_file_system.php
gleimermora: not sure about this one, I can't find this change locally :/

classes/local/store/object_file_system.php
```diff
@@ -316,6 +322,47 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
...
+        $iterator->setFlags(RecursiveDirectoryIterator::SKIP_DOTS);
+        $directories = new ParentIterator($iterator);
+        foreach (new RecursiveIteratorIterator($directories, RecursiveIteratorIterator::CHILD_FIRST) as $dir) {
+            $dirpath = $dir->getPathname();
+            if ($this->is_dir_empty($dirpath)) {
+                rmdir($dirpath);
+            }
+        }
+        // Make sure we keep the 'filedir' dir.
+        if ('filedir' !== basename($rootpath) && $this->is_dir_empty($rootpath)) {
```
brendanheywood: Can you test filedir without this hard coded, eg $CFG->filedir which should be setup earlier?

gleimermora: updated

classes/local/store/object_file_system.php
```diff
@@ -585,7 +632,11 @@ public function delete_object_from_hash($contenthash) {
      */
     public function delete_local_file_from_hash($contenthash) {
         $path = $this->get_local_path_from_hash($contenthash);
-        unlink($path);
+        if (unlink($path)) {
+            // Check grandparent dir for empty dirs.
+            $dirpath = str_replace($contenthash, '', $path) . '..';
```
brendanheywood: this seems awkward, are you just tryng to get the parent dir? I think you want dirname

https://www.php.net/manual/en/function.dirname.php

classes/local/store/object_file_system.php
```diff
@@ -298,11 +302,13 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
 
         if ($initiallocation === OBJECT_LOCATION_DUPLICATED) {
             $localpath = $this->get_local_path_from_hash($contenthash);
-
             if ($this->verify_external_object_from_hash($contenthash)) {
                 $success = unlink($localpath);
 
                 if ($success) {
+                    // Check grandparent dir for empty dirs.
+                    $path = str_replace($contenthash, '', $localpath) . '..';
```
brendanheywood: use dirname?

tests/object_file_system_test.php
```diff
@@ -206,16 +206,66 @@ public function test_delete_object_from_local_by_hash_if_not_local() {
...
         $file = $this->create_duplicated_file();
-        $externalpath = $this->get_external_path_from_hash($file->get_contenthash());
+        $contenthash = $file->get_contenthash();
+        $externalpath = $this->get_external_path_from_hash($contenthash);
         $localpath = $this->get_local_path_from_storedfile($file);
 
         $differentfilepath = __DIR__ . '/fixtures/test.txt';
         copy($differentfilepath, $externalpath);
+        // Get the grandparent path.
+        $grandparentpath = realpath(str_replace($contenthash, '', $localpath) . '..');
```
brendanheywood: dirname?

tests/object_file_system_test.php
```diff
@@ -206,16 +206,66 @@ public function test_delete_object_from_local_by_hash_if_not_local() {
...
 
-        $location = $this->filesystem->delete_object_from_local_by_hash($file->get_contenthash());
+        $location = $this->filesystem->delete_object_from_local_by_hash($contenthash);
 
         $this->assertEquals(OBJECT_LOCATION_EXTERNAL, $location);
         $this->assertFalse(is_readable($localpath));
+        // If grandparent is empty it should be removed.
+        $this->assertFalse(is_readable($grandparentpath));
```
brendanheywood: would file_exists() which can be used for dirs, be better than is_readable?

classes/local/store/object_file_system.php
gleimermora: yes, it doesn't work for `5.6` :disappointed: .
-  `7.0.0 | Added the optional levels parameter.`

tests/object_file_system_test.php
gleimermora: I used the same for consistency, there is not much difference in these cases.

PR #205, opened 2020-01-06
Use admin_settings fields for settings page

ledship:
This pull request makes the plugin use the admin_settings classes for settings.


brendanheywood:
I've added @gleimermora to do a peer review.

There is a couple of things I specifically want to look for:

1) there should be no blocking tests done while building the generic admin tree, as these slow down every admin page across the site. These tests  should either be ajax async, or you can conditionally run them only when on the admin section that matters

2) the first time experience when setting up should still be very minimal, ie we don't want to show admin pages for stores that we can't use

presignedurl_tests.php
```diff
@@ -25,9 +25,9 @@
 
 require_once(__DIR__ . '/../../../config.php');
 require_once(__DIR__ . '/lib.php');
-require_once($CFG->libdir.'/adminlib.php');
+require_once($CFG->dirroot . '/lib/adminlib.php');
 
-admin_externalpage_setup('tool_objectfs_presignedurl_testing');
+admin_externalpage_setup("tool_objectfs_presignedurl_testing");
```
gleimermora: single quotes?

settings.php
```diff
@@ -25,23 +25,100 @@
...
+    if ($client and $client->get_availability() and $PAGE->url->compare(new moodle_url('/admin/settings.php?section=tool_objectfs'))) {
+        $settings = $client->define_client_section($settings, $config);
+    }
+
+    $settings->add(new admin_setting_heading('tool_objectfs/testsettings',
+        new lang_string('settings:testingheader', 'tool_objectfs'), ''));
+
+    $settings->add(new admin_setting_configcheckbox('tool_objectfs/preferexternal',
+        new lang_string('settings:preferexternal', 'tool_objectfs'), '', '', '0', '1'));
 }
```
gleimermora: new line?

object_status.php
```diff
@@ -49,7 +48,4 @@
     echo $output->render_objectfs_report($report);
 }
 
-echo $output->footer();
-
-
-
+echo $output->footer();
```
gleimermora: new line?

missing_files.php
```diff
@@ -25,14 +25,17 @@
 
 require_once(__DIR__ . '/../../../config.php');
 require_once(__DIR__ . '/lib.php');
-require_once($CFG->libdir.'/adminlib.php');
+require_once($CFG->dirroot . '/lib/adminlib.php');
+
+admin_externalpage_setup("tool_objectfs_missing_files");
```
gleimermora: single quotes?

classes/local/store/azure/client.php
```diff
@@ -325,31 +325,31 @@ protected function get_exception_details(ServiceException $exception) {
      * Container name.
      * Shared Access Signature.
      *
-     * @param $mform
+     * @param $settings
```
gleimermora: add type?

ledship: Resolved in commit b5f595d

missing_files.php
ledship: Resolved in commit b5f595d

object_status.php
ledship: Resolved in commit b5f595d

settings.php
ledship: Resolved in commit b5f595d

presignedurl_tests.php
ledship: Resolved in commit b5f595d

classes/local/store/s3/client.php
```diff
@@ -292,35 +292,33 @@ protected function get_exception_details($exception) {
...
         }
-        return $mform;
     }
 
-    public function define_client_section($mform, $config) {
 
-        $mform->addElement('header', 'awsheader', get_string('settings:aws:header', 'tool_objectfs'));
-        $mform->setExpanded('awsheader');
+    /**
+     * @param $settings admin_settingpage
```
gleimermora: suggested syntax for the `param` tag:
- `@param [Type] [name] [<description>]`

settings.php
```diff
@@ -25,23 +25,98 @@
...
+            new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), '', 10 * MINSECS));
+
+        $settings->add(new filesize('tool_objectfs/presignedminfilesize',
+            new lang_string('settings:presignedurl:presignedminfilesize', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:presignedminfilesize_help', 'tool_objectfs'), 0));
+    }
+
+    $client = tool_objectfs_get_client($config);
+    if ($client and $client->get_availability() and $PAGE->url->compare(new moodle_url('/admin/settings.php?section=tool_objectfs'))) {
```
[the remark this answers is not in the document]
ledship: Resolved in 37702e7

classes/local/store/s3/client.php
ledship: Resolved in 37702e7

classes/local/store/s3/client.php
```diff
@@ -292,35 +292,33 @@ protected function get_exception_details($exception) {
         return $details;
     }
 
-    public function define_amazon_s3_check($mform, $testdelete = true) {
-        global $OUTPUT;
-        $connection = false;
-
+    public function define_amazon_s3_check($testdelete = true) {
+        global $SESSION;
+        $SESSION->notifications = [];
```
brendanheywood: This feels wrong to me, wont this get run on every admin page when the admin tree is loaded?

settings.php
```diff
@@ -25,23 +25,98 @@
...
 
-    $externalpage = new admin_externalpage('tool_objectfs_settings',
-                                            get_string('pluginname', 'tool_objectfs'),
-                                            new moodle_url('/admin/tool/objectfs/index.php'));
+    $ADMIN->add('tools', new admin_externalpage('tool_objectfs_presignedurl_testing',
+        get_string('presignedurl_testing:page', 'tool_objectfs'),
+        new moodle_url('/admin/tool/objectfs/presignedurl_tests.php')));
 
-    $ADMIN->add('tools', $externalpage);
+    $ADMIN->add('tools', new admin_externalpage('tool_objectfs_object_status',
```
brendanheywood: Shouldn't these stay / or be moved into the reports admin tab?

classes/local/store/s3/client.php
ledship: No, because of the check made on line 113 of settings.php which checks if the current page is the settings page. I added that check to stop it from trying to test the connection when it is on any other admin page.

settings.php
```diff
@@ -25,23 +25,98 @@
...
+
+        $settings->add(new admin_setting_configcheckbox('tool_objectfs/enablepresignedurls',
+            new lang_string('settings:presignedurl:enablepresignedurls', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:enablepresignedurls_help', 'tool_objectfs'), '', '0', '1'));
+
+        $settings->add(new admin_setting_configduration('tool_objectfs/expirationtime',
+            new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), '', 10 * MINSECS));
+
+        $settings->add(new filesize('tool_objectfs/presignedminfilesize',
```
gleimermora: not sure about this class `filesize`,  could you post the path to it?

ledship: Opps, I thought I removed all references to that in one of the last commits. It is 3.8 only and I didn't realise that when I initially added it.

ledship: Resolved in 654263b

settings.php
ledship: Yep, didn't catch that, done in latest commit.

settings.php
```diff
@@ -25,23 +25,98 @@
...
+            new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), '', 10 * MINSECS));
+
+        $settings->add(new admin_setting_configtext('tool_objectfs/presignedminfilesize',
+            new lang_string('settings:presignedurl:presignedminfilesize', 'tool_objectfs'),
+            new lang_string('settings:presignedurl:presignedminfilesize_help', 'tool_objectfs'), 0, PARAM_INT));
+    }
+
+    $client = tool_objectfs_get_client($config);
+    if ($client && $client->get_availability() && $PAGE->url->compare(new moodle_url('/admin/settings.php?section=tool_objectfs'))) {
```
brendanheywood: This feels like it should be split in two, if the client is available then I think you should show define_client_section

But then inside that, only if you are in ?section=tool_objectfs should you do the extra step of the 'define_client_check'

ie I think define_client_check needs to be called from settings.php and have a consistent api, ie rename define_amazon_s3_check

classes/local/store/s3/client.php
brendanheywood: Ah see my comments here https://github.com/catalyst/moodle-tool_objectfs/pull/205/files#r364041364

settings.php
ledship: What about the argument supplied to define_amazon_s3_check? Remove it and the condition check on line 250? Or specifically check in settings.php if it is a DO filesystem and then supply false in the arguments?

brendanheywood: might be worth the DO one overriding it and adding the second check instead of one shared method with a code fork

classes/local/store/digitalocean/client.php
```diff
@@ -53,10 +53,32 @@ public function set_client($config) {
         ));
     }
 
-    public function define_client_section($mform, $config) {
+    public function define_client_check($client) {
```
gleimermora: this seems duplicated from the base class `object_client_base`

ledship: `define_client_check` overrides `object_client_base`'s function as it needs to supply false to the test permissions function

gleimermora: you could set a property inside `object_client_base` with default value of `true` then override this value in `digitalocean/client` and set it to false.
Then you could:
```
// Check permissions if we can connect.
$permissions = $client->test_permissions($this->testdelete);
```
Also for consistency it would be a good idea to update the signature of `test_permissions` for the azure/swift clients, perhaps it should be included in the `interface`?

brendanheywood:
thanks @ledship and @gleimermora this looks great

Issue #169, opened 2019-08-07
When using signed urls with AWS, have an option to use S3 signing vs CloudFront signing


brendanheywood:
Not sure what we need here in terms of extra api keys

PR #206, opened 2020-01-06
Add cloudfront signing option for s3 store

nickmccutcheon:
This is a pull request for issue #169


classes/local/form/settings_form.php
```diff
@@ -168,23 +168,23 @@ public function define_client_selection($mform, $config) {
     }
 
     public function define_presignedurl_section($mform) {
-        global $OUTPUT;
+        global $OUTPUT, $config;
```
brendanheywood: This will need some rework now to convert to admin settings instead of formslib

PR #208, opened 2020-01-08
Use proper filedir property

gleimermora:


classes/task/delete_local_empty_directories.php
```diff
@@ -53,9 +53,6 @@ public function execute() {
             return;
         }
         $filesystem = new $config->filesystem();
-        if (!isset($CFG->filedir)) {
-            $CFG->filedir = $CFG->dataroot . '/filedir';
-        }
-        $filesystem->delete_empty_folders($CFG->filedir);
+        $filesystem->delete_empty_folders($CFG->dataroot . '/filedir');
```
brendanheywood: The filesystem knows it's own filedir, it make me think this should just make the path optional and let the method handle it

gleimermora: updated

PR #210, opened 2020-01-10
Clean up delete_local_empty_directories task

gleimermora:


tests/object_file_system_test.php
```diff
@@ -229,16 +227,16 @@ public function test_delete_object_from_local_by_hash_if_can_verify_external_obj
     public function delete_empty_folders_provider() {
         return [
             [
-                ['/d1', '/d2'], ['/file1', '/file2'], [true, true],  true,
+                ['/d1', '/d2'], ['/file1', '/file2'], [true, true],  true, 0,
```
brendanheywood: It's not obvious to me what these tests do, the actual directories and files will actually be with these tests, and also what the end state will be. Can I suggest you make this data provider give 3 things:

1) A prior state
2) A file to be deleted
3) A list of dirs to confirm are also gone as a result

And in a very readable format like this:

```
before => "/aa/bb/aabbcc
/aa/bb/aabbdd
/aa/cc/aaccee"
delete => "/aa/bb/aabbdd"
gone => []

before => "/aa/bb/aabbcc
/aa/cc/aaccee"
delete => "/aa/bb/aabbcc"
gone => [/aa/bb/]

before => "/aa/cc/aaccee"
delete => "/aa/cc/aaccee"
gone => [/aa/]

```

Issue #193, opened 2019-11-03
filedir local files double check cron task


brendanheywood:
We know that in some cases there are files being left in the filedir but are incorrectly being marked as in object storage. We need to track down why that is happening, but as a second set of protection we should run a cron task which ignores the data base and works purely from the filesystem and double checks these files.

*Report*
* [x] have a function which finds filedir size and count
* [x] run this from cron and save to plugin config
* [x] add these numbers to the normal status report
* [x]  ~~have a button to update this which also unlocks the session~~ just link to the scheduled task for this

*Fixing task*
* [ ] recurse through the filedir and for every file it finds check if objectfs thinks it is either in the remote state or the missing state, and if so correct it's metadata
* [ ] for most sites this list of files should be very small and we could possibly do a full scan. But possibly it will be so large we need to stop and start it. In which case we should work through the files in alphabetic order of their content hash, and store the has of where we are up to.

brendanheywood:
Note: putting this issue on ice, the root cause problem is different to what we expected and is due to a mdl_files being out of sync with filedir, and not objectfs being out of sync with either of those. So this patch doesn't address the issue we want to fix. The root cause it likely either a core bug, nor not even a bug but just legacy files floating around from the dark ages.

brendanheywood:
Closing this is done here:

https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_404_STABLE/classes/task/reconcile_filedir.php

PR #211, opened 2020-01-16
Add task to sync file location from filedir #193

gleimermora:
https://github.com/catalyst/moodle-tool_objectfs/issues/193

### DIFF
--- a/classes/local/object_manipulator/checker_filedir.php
+++ b/classes/local/object_manipulator/checker_filedir.php
@@ -25,8 +25,6 @@
 
 namespace tool_objectfs\local\object_manipulator;
 
-use coding_exception;
-use dml_exception;
 use stdClass;
 use tool_objectfs\local\store\object_file_system;
 use tool_objectfs\log\aggregate_logger;
@@ -37,105 +35,57 @@ require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
 
 class checker_filedir extends manipulator {
 
-    /**
-     * @var array $totalfiles
-     */
-    private $totalfiles = [];
-
-    /**
-     * @var array $files
-     */
-    private $files = [];
+    /** @var bool $timeexceeded */
+    public $timeexceeded = false;
 
     /**
      * sync_filedir constructor.
      * @param object_file_system $filesystem
      * @param $config
      * @param aggregate_logger $logger
-     * @throws dml_exception
      */
     public function __construct(object_file_system $filesystem, $config, aggregate_logger $logger) {
         parent::__construct($filesystem, $config);
         $this->logger = $logger;
         // Inject our logger into the filesystem.
         $this->filesystem->set_logger($this->logger);
-        $this->files = $this->filter_files();
     }
 
     /**
      * @return string
      */
     protected function get_query_name() {
-        return 'get_checker_filedir_candidates';
+        return '';
     }
 
     /**
      * @return string
-     * @throws coding_exception
-     * @throws dml_exception
      */
     protected function get_candidates_sql() {
-        global $DB;
-        list($inclause, ) = $DB->get_in_or_equal($this->files);
-        return 'SELECT id, contenthash
-                FROM {tool_objectfs_objects}
-                WHERE location > ?
-                    AND contenthash ' . $inclause;
+        return '';
     }
 
     /**
      * @return array
-     * @throws coding_exception
-     * @throws dml_exception
      */
     protected function get_candidates_sql_params() {
-        global $DB;
-        list(, $params) = $DB->get_in_or_equal($this->files);
-        array_unshift($params, OBJECT_LOCATION_DUPLICATED);
-        return $params;
+        return [];
     }
 
     /**
-     * @param $objectrecord
+     * @param stdClass $objectrecord
      * @return int
      */
     protected function manipulate_object($objectrecord) {
-        return OBJECT_LOCATION_DUPLICATED;
+        return $this->filesystem->get_object_location_from_hash($objectrecord->contenthash);
     }
 
     /**
-     * Set the las file name processed or clean the records if all have been processed.
-     * @param stdClass $lastprocessed
-     * @throws dml_exception
+     * Set the las file name processed after exceeding the max exec time.
+     * @param string $lastprocessed
      */
-    protected function set_last_processed($lastprocessed) {
-        $lastcontenthash = end($this->files);
-        if (end($this->totalfiles) === $lastcontenthash) {
-            set_config('lastprocessed', '', 'tool_objectfs');
-            return;
-        }
-        $obj = new stdClass();
-        $obj->contenthash = $lastcontenthash;
-        if (!empty($lastprocessed->contenthash)) {
-            $obj->contenthash = $lastprocessed->contenthash;
-        }
-        set_config('lastprocessed', $obj->contenthash, 'tool_objectfs');
-    }
-
-    /**
-     * Get the file batch starting from the last processed file name and limited by the batchsize value.
-     * @return array
-     * @throws dml_exception
-     */
-    private function filter_files() {
-        $lastprocessed = get_config('tool_objectfs', 'lastprocessed');
-        $this->totalfiles = $this->filesystem->get_filenames_from_dir();
-        $files = $this->totalfiles;
-        if (!empty($lastprocessed)) {
-            $files = array_filter($this->totalfiles, function ($name) use ($lastprocessed) {
-                return $name > $lastprocessed ? true : false;
-            });
-        }
-        return array_slice($files, 0, $this->batchsize);
+    public function set_last_processed($lastprocessed) {
+        $this->timeexceeded = true;
+        set_config('lastprocessed', $lastprocessed, 'tool_objectfs');
     }
 }
--- a/classes/local/object_manipulator/checker_filedir.php
+++ b/classes/local/object_manipulator/checker_filedir.php
@@ -35,7 +35,7 @@ defined('MOODLE_INTERNAL') || die();
 
 require_once($CFG->dirroot . '/admin/tool/objectfs/lib.php');
 
-class sync_filedir extends manipulator {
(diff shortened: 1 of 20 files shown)


classes/local/object_manipulator/manipulator.php
```diff
@@ -188,5 +197,13 @@ public static function setup_and_run_object_manipulator($manipulatorclassname) {
...
         $candidatehashes = $manipulator->get_candidate_objects();
         $manipulator->execute($candidatehashes);
+        $manipulator->set_last_processed($manipulator->lastprocessed);
+    }
+
+    /**
+     * @param stdClass $lastprocessed
+     */
+    protected function set_last_processed($lastprocessed) {
+        return;
```
brendanheywood: What is the point of this?

db/install.xml
```diff
@@ -28,5 +28,15 @@
         <KEY NAME="primary" TYPE="primary" FIELDS="id"/>
       </KEYS>
     </TABLE>
+    <TABLE NAME="tool_objectfs_sync_location" COMMENT="Object sync location for objectfs">
```
brendanheywood: Why do you need a new table?

classes/local/object_manipulator/manipulator.php
gleimermora: I wanted to avoid the 
```
if ($manipulator instanceof sync_filedir:class)
    $manipulator->set_last_processed(...
```

classes/local/object_manipulator/sync_filedir.php
```diff
@@ -0,0 +1,156 @@
...
+        array_unshift($params, OBJECT_LOCATION_DUPLICATED);
+        return $params;
+    }
+
+    /**
+     * @param $objectrecord
+     * @return int
+     */
+    protected function manipulate_object($objectrecord) {
+        return OBJECT_LOCATION_DUPLICATED;
```
brendanheywood: I'm not sure this is safe. Lets say the file is on local disc, and is not remote.

If you set it as duplicated then a later cron job might delete it from local, thinking it is safely remote when it isn't.

So instead it needs to check where it is remote or not, and either set it to either LOCAL or DUPLICATED, there is existing code for this, see example:

https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/object_manipulator/checker.php#L68-L71

classes/local/store/object_file_system.php
```diff
@@ -354,6 +355,28 @@ public function delete_empty_folders($rootpath = null) {
         return $deletedircount;
     }
 
+    /**
+     * Get the name of the files from a specific dir.
+     * @param string $dir
+     * @return array
+     */
+    public function get_filenames_from_dir($dir = '') {
+        if (empty($rootpath)) {
```
brendanheywood: typo? $rootpath vs $dir?

classes/local/store/object_file_system.php
```diff
@@ -354,6 +355,28 @@ public function delete_empty_folders($rootpath = null) {
...
+            $dir = $this->filedir;
+        }
+        $iterator = new RecursiveDirectoryIterator($dir);
```
brendanheywood: The RecursiveDirectoryIterator scares me a little because I can't see how it works. Is it breadth first or depth first? When does it stop? What order will it process things in?

Lets assume we have a fresh moodle instance from a new client, and there is 10's of millions of local files here because we haven't synced anything to the remote store yet. This task will be trying to grab a full complete recursive listing and stick it all in memory, and then sort it all in memory, and then do something with it.

This all seems like a scaling issue waiting to happen, and testing this on a current prod site which has already been mostly migrated isn't a valid test.

Unless we can guarantee the way the RecursiveDirectoryIterator works, I'd prefer to see an approach with manual recursion so we have full control over how it all works. ie I'd get a root dir listing (max 256 items), then sort it, then drill into the next dir, get a listing, then sort it, then process files 1 by one. Then if we get to a max execution time we mark the file we are at and stop, and at no point did we ever attempt to get a full listing. The next run we would pick up from where we left off.

tests/tasks_test.php
```diff
@@ -61,6 +61,7 @@ public function test_run_scheduled_tasks() {
             'push_objects_to_storage',
             'recover_error_objects',
             'check_objects_location',
+            'sync_filedir_location',
```
brendanheywood: This new task is broadly the same as the check_objects_location, but one is looking for orphaned db records, and the other is looking for orphaned files. I'm wondering wether we should name them similarly so this relationship is more obvious.

eg check_objects_location vs check_filedir_location

or even renaming them both:

check_location_db and check_location_local_filedir

db/tasks.php
```diff
@@ -89,5 +89,14 @@
         'dayofweek' => '*',
         'month'     => '*'
     ),
+    [
+        'classname' => 'tool_objectfs\task\sync_filedir_location',
+        'blocking'  => 0,
+        'minute'    => '0',
+        'hour'      => '*/8',
```
brendanheywood: Can you run this just once a night, say 2am

db/install.xml
gleimermora: removed

classes/local/store/object_file_system.php
gleimermora: nice catch :+1:

tests/tasks_test.php
gleimermora: renamed to check_filedir_location

db/tasks.php
gleimermora: updated

classes/local/store/object_file_system.php
brendanheywood: Adding the flags to this like the other one should be perfect

classes/local/object_manipulator/checker_filedir.php
```diff
@@ -0,0 +1,91 @@
...
+        parent::__construct($filesystem, $config);
+        $this->logger = $logger;
+        // Inject our logger into the filesystem.
+        $this->filesystem->set_logger($this->logger);
+    }
+
+    /**
+     * @return string
+     */
+    protected function get_query_name() {
```
brendanheywood: It feels wrong to have these 3 empty methods here. The alternative is splitting manipulator into an 'sql_manipulator' for the other 5 and only this one extends manipulator directly

At the very least add some comments saying why they are redundant.

classes/local/object_manipulator/checker_filedir.php
```diff
@@ -0,0 +1,91 @@
...
+    protected function manipulate_object($objectrecord) {
+        return $this->filesystem->get_object_location_from_hash($objectrecord->contenthash);
+    }
+
+    /**
+     * Set the las file name processed after exceeding the max exec time.
+     * @param string $lastprocessed
+     */
+    public function set_last_processed($lastprocessed) {
+        $this->timeexceeded = true;
```
brendanheywood: Is this intentional? doesn't look right

classes/local/object_manipulator/manipulator.php
```diff
@@ -25,6 +25,8 @@
 
 namespace tool_objectfs\local\object_manipulator;
 
+use stdClass;
```
brendanheywood: is this actually used?

classes/local/store/object_file_system.php
```diff
@@ -45,10 +47,14 @@
...
 abstract class object_file_system extends \file_system_filedir {
 
+    const MAX_ROOT_DIRS = 256;
+    const MAX_FILES_PER_DIR_BATCH = 1000;
+
     private $externalclient;
     private $preferexternal;
     private $deleteexternally;
     private $logger;
+    private $found = false;
```
brendanheywood: Unused?

classes/local/store/object_file_system.php
```diff
@@ -365,29 +371,96 @@ public function delete_empty_folders($rootpath = null) {
             $emsg = $e->getMessage();
             $path = substr($emsg, strrpos($emsg, '(') + 1, strrpos($emsg, ')') - 1);
             $msg = substr($emsg, strrpos($emsg, ':') + 2);
-            $message = "Delete empty dir failed: $path $msg {$this->get_path_owner_info($path)}";
+            $message = "Delete empty dir failed: $path $msg";
```
brendanheywood: Did this leak from the other patch?

gleimermora: yes, but I'll get removed if we merged https://github.com/catalyst/moodle-tool_objectfs/pull/220

classes/local/object_manipulator/checker_filedir.php
gleimermora: I'll address this in a another pr

classes/local/object_manipulator/checker_filedir.php
gleimermora: updated

classes/local/object_manipulator/manipulator.php
gleimermora: removed

classes/local/store/object_file_system.php
```diff
@@ -345,6 +354,71 @@ public function delete_empty_dirs($rootpath = '') {
...
+     * Get the name of the file names from a specific dir.
+     * @param string $dir
+     * @param string $lastprocessedhash
+     * @return Generator
+     */
+    public function get_filenames_from_dir($dir = '', $lastprocessedhash = '') {
+        if (empty($dir)) {
+            $dir = $this->filedir;
+        }
+        $flags = (RecursiveDirectoryIterator::SKIP_DOTS | RecursiveDirectoryIterator::FOLLOW_SYMLINKS);
```
brendanheywood: Is the order of this strictly alphabetical?

brendanheywood:
Note: putting this PR on ice, the root cause problem is different to what we expected and is due to a mdl_files being out of sync with filedir, and not objectfs being out of sync with either of those. So this patch doesn't address the issue we want to fix. The root cause it likely either a core bug, nor not even a bug but just legacy files floating around from the dark ages.

PR #216, opened 2020-01-21
Add try/catch to avoid tasks to failed and log the error

gleimermora:


brendanheywood:
I think we should close this until we can reproduce the issue properly

classes/local/store/object_file_system.php
```diff
@@ -334,7 +340,7 @@ public function delete_empty_folders($rootpath = null) {
             return 0;
         }
         $deletedircount = 0;
-        $iterator = new RecursiveDirectoryIterator($rootpath);
+        $iterator = new RecursiveDirectoryIterator($rootpath, RecursiveDirectoryIterator::FOLLOW_SYMLINKS);
```
brendanheywood: what is this about? there should not be any symlinks in filedir

gleimermora: just ccurious why goes to `/var/lib/sitedata/filedir/16/fb` instead of `/var/lib/sitedata/ecs_marist_moodle_stage/filedir`

Issue #192, opened 2019-11-03
When we remove a file from filedir we should also cleanup the parent dir if it is empty


brendanheywood:
These empty dirs make dir listings and scans much slower

PR #220, opened 2020-01-24
Refactor delete empty dirs #192

gleimermora:


classes/local/store/object_file_system.php
```diff
@@ -325,87 +325,21 @@ public function delete_object_from_local_by_hash($contenthash, $objectsize = 0)
...
-        if ($handle = opendir($foldername)) {
-            while (false !== ($file = readdir($handle))) {
-                if ($file !== '.' && $file !== '..') {
-                    closedir($handle);
-                    return false;
-                }
-            }
-            closedir($handle);
+        $empty = true;
+        foreach (glob($rootpath . DIRECTORY_SEPARATOR . '*') as $path) {
```
brendanheywood: Could we also use GLOB_ONLYDIR here?

Issue #221, opened 2020-02-10
$PAGE->set_url() is not set


golenkovm:
```
PHP Notice:  This page did not call $PAGE->set_url(...). Using * line 599 of /lib/pagelib.php: call to debugging()
* line 815 of /lib/pagelib.php: call to moodle_page->magic_get_url()
* line 120 of /admin/tool/objectfs/settings.php: call to moodle_page->__get()
* line 76 of /lib/classes/plugininfo/tool.php: call to include()
* line 557 of /admin/settings/plugins.php: call to core\plugininfo\tool->load_settings()
* line 8073 of /lib/adminlib.php: call to require()
* line 8094 of /lib/adminlib.php: call to admin_get_root()
* line 196 of /admin/cli/upgrade.php: call to admin_apply_default_settings()
 in /var/www/clients/catalyst/totara/lib/weblib.php on line 3349
```

golenkovm:
Hey @ledship 
Could you have a look at this issue? I believe it originates from https://github.com/catalyst/moodle-tool_objectfs/commit/cd3194d7d032a6600901a941e4dc11ff3635e4b2

azrek:
PR https://github.com/catalyst/moodle-tool_objectfs/pull/222

azrek:
This PR does not fix the broken menu. It only fixes the error when upgrading through CLI.

PR #222, opened 2020-02-10
Issue #221: fix for PAGE->set_url()

azrek:
Also fix error when cil upgrade is run and
checking for connection. Session is not available
to display notification.

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -117,7 +117,7 @@ if ($ADMIN->fulltree) {
     if ($client && $client->get_availability()) {
         $settings = $client->define_client_section($settings, $config);
 
-        if ($PAGE->url->compare(new moodle_url('/admin/settings.php?section=tool_objectfs'))) {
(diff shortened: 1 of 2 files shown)


settings.php
```diff
@@ -115,6 +115,7 @@
 
     $client = tool_objectfs_get_client($config);
     if ($client && $client->get_availability()) {
+        $PAGE->set_url('/admin/settings.php?section=tool_objectfs');
```
brendanheywood: This doesn't look right, the $PAGE->url->compare below is to detect when we are on this page, vs on the admin search page. If $PAGE is not set in the cli then instead you should only run define_client_check() below if $PAGE is set, and the url compare is good. That probably makes the session check above redundant too but it's safe to keep either way

azrek: You're absolutely correct @brendanheywood 

Pushed a change.

PR #223, opened 2020-02-10
Add objectfs settings link to admin tools menu on plugins page

ledship:


settings.php
```diff
@@ -35,6 +35,9 @@
 
 if (!$ADMIN->locate('tool_objectfs')) {
```
brendanheywood: Do you understand why this if exists? This seems very non standard and I suspect might be the root cause of the weirdness

PR #225, opened 2020-02-11
Check class exists before using it

azrek:
During the render of the Object Status report,
check the class tool_task\run_from_cli exists.
It did not come into Moodle until version 3.5

And, default the runnable task to false.


azrek:
I'm not sure if this is the correct behaviour - i.e. runnable = false.

Or, if we should create the link to allow the user to run the task.

brendanheywood:
This looks good ot me, that run task from gui is a new feature in moodle which hasn't landed in totara

Issue #226, opened 2020-02-11
Unit tests failing on empty grandparent directories


azrek:
1) tool_objectfs\tests\object_file_system_testcase::test_delete_object_from_local_by_hash_if_can_verify_external_object
Failed asserting that true is false.

/var/www/site/admin/tool/objectfs/tests/object_file_system_test.php:221
/var/www/site/lib/phpunit/classes/base_testcase.php:600
/var/www/site/lib/phpunit/classes/advanced_testcase.php:68

To re-run:
 vendor/bin/phpunit tool_objectfs\tests\object_file_system_testcase admin/tool/objectfs/tests/object_file_system_test.php

FAILURES!
Tests: 1425, Assertions: 15951, Failures: 1, Skipped: 5.

PR #228, opened 2020-02-11
Issue #226: Unit tests are failing checking empty directories

azrek:
Needed to clear out old directory structures and recreate
according to test.

### DIFF
--- a/tests/object_file_system_test.php
+++ b/tests/object_file_system_test.php
@@ -291,7 +291,7 @@ class object_file_system_testcase extends tool_objectfs_testcase {
         }
         $this->filesystem->delete_empty_dirs($testdir);
         foreach ($dirs as $key => $dir) {
-             $this->assertEquals($expectedparentreadable[$key], is_readable($testdir . $dir));
+            $this->assertEquals($expectedparentreadable[$key], is_readable($testdir . $dir));
         }
         $this->assertEquals($expectedgrandparentpathreadable, is_readable($testdir));
     }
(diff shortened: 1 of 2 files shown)


azrek:
hi @brendanheywood  / @gleimermora 

I'm not sure If this is the correct way to implement the recursive deletion of the existing test directory.

But, the issue was due to the directory and its children not getting delete before the test is run. It's a matter of cleaning up the structure for the test.

I'll be guided by you.

Thanks,
Ghada

tests/object_file_system_test.php
```diff
@@ -291,9 +293,11 @@ public function test_delete_empty_folders(
         }
         $this->filesystem->delete_empty_dirs($testdir);
         foreach ($dirs as $key => $dir) {
-             $this->assertEquals($expectedparentreadable[$key], is_readable($testdir . $dir));
+            $this->assertEquals($expectedparentreadable[$key], is_readable($testdir . $dir));
         }
         $this->assertEquals($expectedgrandparentpathreadable, is_readable($testdir));
+        // clean up.
+        $this->rrmdir($testdir);
```
gleimermora: if the issue is `resetAfterTest` is not working I think it is safe to do:
```
shell_exec('rm -rf ' .  $testdir);
```
I've tested it locally(totara:CatalystLearn12-3)

azrek: Pushed changes to clear filedir after each test.

Issue #231, opened 2020-02-13
Relative URLs in the scorm are pointing to S3


jackson-catalyst:
In addition to the reported bug #167, we manually changed the mime-type on S3 launched the scorm and noticed that all relative URLs in the scorm are pointing to S3

Suggested fixes are to not server HTML and other files that could have relative links from S3.

brendanheywood:
Maybe duplicate https://github.com/catalyst/moodle-tool_objectfs/issues/227

PR #232, opened 2020-02-14
Test catalyst #231 - Test file for relative URLs

jackson-catalyst:
This is a patch to test S3 URL signing for file (e.g. SCORMS, etc...) with relative URLs. It currently fails.


brendanheywood:
Closing this, @gleimermora is working on both the tests for this and the fix and has some deeper tests for a range of file types in progress

Issue #233, opened 2020-02-21
Set default value for `tool_objectfs/expirationtime` setting


golenkovm:
Objectfs doesn't have a default value for `tool_objectfs/expirationtime` setting (One of Pre-Signed URLs Settings)
https://github.com/catalyst/moodle-tool_objectfs/blob/b440120dfdc662d7c1b51ba605a103fee777a48b/settings.php#L103-L105

After upgrade Moodle/Totara asks it to be defined and it might be confusing for admins:

![image](https://user-images.githubusercontent.com/22066290/74995032-c9c8bd80-54a3-11ea-8151-39f38aabafd0.png)

PR #234, opened 2020-03-02
Fix default value for tool_objectfs/expirationtime setting 

golenkovm:
Fix for #233

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -102,7 +102,7 @@ if ($ADMIN->fulltree) {
 
         $settings->add(new admin_setting_configduration('tool_objectfs/expirationtime',
             new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
-            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), 10, MINSECS));
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), 2 * HOURSECS, MINSECS));
 
         $settings->add(new admin_setting_configtext('tool_objectfs/presignedminfilesize',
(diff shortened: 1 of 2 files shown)


settings.php
```diff
@@ -102,7 +102,7 @@
 
         $settings->add(new admin_setting_configduration('tool_objectfs/expirationtime',
             new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
-            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), '', 10 * MINSECS));
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), 10, MINSECS));
```
gleimermora: I think it should be `10 * MINSECS` if we want `10 mins`

brendanheywood: This should be minutes, and also 10 is waaaaayyy to short. This needs to be around the same as the session expiry.

$CFG->sessiontimeout now defaults to 8 hours, so lets make this expiry at least 2 hours

settings.php
```diff
@@ -102,7 +102,7 @@
 
         $settings->add(new admin_setting_configduration('tool_objectfs/expirationtime',
             new lang_string('settings:presignedurl:expirationtime', 'tool_objectfs'),
-            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), '', 10 * MINSECS));
+            new lang_string('settings:presignedurl:expirationtime_help', 'tool_objectfs'), 2 * HOURSECS, MINSECS));
```
golenkovm: Let's set $defaultunit to HOURSECS as well

gleimermora: updated

PR #235, opened 2020-03-04
WR330605 settings page broken on moodle version lower than 3.1

gleimermora:


classes/local/store/object_client_base.php
```diff
@@ -108,20 +108,39 @@ public function define_client_check($client) {
...
+    private function notification($message, $type = 'success') {
+        global $CFG, $OUTPUT;
+        $methodsmap = [
+            'success' => 'notify_success',
+            'warning' => 'notify_problem',
+            'error' => 'notify_problem',
+        ];
+        if (!empty($CFG->branch) && $CFG->branch < '31' && isset($methodsmap[$type])) {
+            $method = $methodsmap[$type];
+            $OUTPUT->$method($message);
```
brendanheywood: why not just call

public function notification($message, $type = null) {

$OUTPUT->motification($message, $type); ?

classes/local/store/object_client_base.php
```diff
@@ -108,20 +108,39 @@ public function define_client_check($client) {
         $SESSION->notifications = [];
```
brendanheywood: I don't think you need this anymore, or the line 103 as well

classes/local/store/object_client_base.php
gleimermora: `$OUTPUT->$method` is deprecated on versions 3.1+ and it throws an exception.

gleimermora:
new branch: https://github.com/catalyst/moodle-tool_objectfs/pull/237

Issue #213, opened 2020-01-18
update_object_records grabs the record again in cases where it doesn't need to


brendanheywood:
Here we have the object:

https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/object_manipulator/manipulator.php#L123

We should just pass the object through and conditionally save a second db lookup here:

https://github.com/catalyst/moodle-tool_objectfs/blob/master/lib.php#L39-L47

This effects every manipulator, this should massively reduce the db calls against the object table

PR #238, opened 2020-03-06
Objectfs - Save extra query when manipulator class updates and object…

gleimermora:
… record #213

### DIFF
--- a/classes/local/object_manipulator/checker.php
+++ b/classes/local/object_manipulator/checker.php
@@ -49,15 +49,22 @@ class checker extends manipulator {
         return 'get_check_candidates';
     }
 
+    /**
+     * @return string
+     */
     protected function get_candidates_sql() {
-        $sql = 'SELECT f.contenthash
+        return 'SELECT MAX(f.id) AS fid,
+                       o.id,
+                       f.contenthash,
+                       MAX(f.filesize) AS filesize
                   FROM {files} f
              LEFT JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
                  WHERE f.filesize > 0
                    AND o.location is NULL
-              GROUP BY f.contenthash';
-
-        return $sql;
+              GROUP BY o.id,
+                       f.contenthash,
+                       f.filesize,
+                       o.location';
     }
 
     protected function get_candidates_sql_params() {
--- a/classes/local/object_manipulator/manipulator.php
+++ b/classes/local/object_manipulator/manipulator.php
@@ -25,6 +25,8 @@
 
 namespace tool_objectfs\local\object_manipulator;
 
+use dml_exception;
+
 defined('MOODLE_INTERNAL') || die();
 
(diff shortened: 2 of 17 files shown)


classes/local/object_manipulator/checker.php
```diff
@@ -50,12 +50,16 @@ protected function get_query_name() {
     }
 
     protected function get_candidates_sql() {
-        $sql = 'SELECT f.contenthash
+        $sql = 'SELECT f.contenthash,
```
brendanheywood: purely for consistency I think the column order here should match the others ie id, hash then size

lib.php
```diff
@@ -36,36 +36,55 @@
...
-    global $DB;
 
-    $newobject = new \stdClass();
+/**
+ * @param string $contenthash
+ * @param int $newlocation
+ * @return stdClass
+ * @throws dml_exception
+ */
+function update_object_record($contenthash, $newlocation) {
```
brendanheywood: this is good but I found the name a little misleading, as I would have expected this to be the one which accepted an object. How about instead 'update_object_by_hash'

classes/local/object_manipulator/checker.php
gleimermora: updated

lib.php
gleimermora: updated to update_object_by_hash

classes/local/object_manipulator/manipulator.php
```diff
@@ -25,6 +25,8 @@
 
 namespace tool_objectfs\local\object_manipulator;
 
+use dml_exception;
```
Peterburnett: This doesnt seem to be necessary. Mightve been auto pulled in by the IDE

tests/tool_objectfs_testcase.php
```diff
@@ -18,6 +18,9 @@
 
 defined('MOODLE_INTERNAL') || die();
 
+use dml_exception;
```
Peterburnett: These also probably shouldnt be here

classes/local/object_manipulator/checker.php
```diff
@@ -49,15 +49,22 @@ protected function get_query_name() {
         return 'get_check_candidates';
     }
 
+    /**
+     * @return string
+     */
     protected function get_candidates_sql() {
-        $sql = 'SELECT f.contenthash
+        return 'SELECT MAX(f.id) AS fid,
```
Peterburnett: After grepping the Codebase, the fid is never used, as you have swapped over to o.id as the primary identifier. Is the fid still needed? I could be missing something

classes/local/object_manipulator/manipulator.php
gleimermora: it just looks a bit cleaner with the `use` statement. :)

classes/local/object_manipulator/checker.php
gleimermora: `fid` is not referenced, I just want to keep it for consistency.

Issue #64, opened 2017-03-29
rethink get_candidate_object


ilya-catalyst:
We should probably call it before creating fillesystem. Because if there are no candidates we don't need filesystem at all. Make it static function?

kenneth-hendricks:
Probably make it a separate class - its violating single responsibility right now.

Should refactor out into object_finders. Will make tests cleaner as well

PR #240, opened 2020-03-12
Manipulator classes refactoring  #64

gleimermora:


classes/local/object_manipulator/candidates/candidates_factory.php
```diff
@@ -0,0 +1,72 @@
...
+        recoverer::class => recoverer_candidates::class,
+    ];
+
+    /**
+     * candidates_factory constructor.
+     * @param string $manipulator
+     * @param stdClass $config
+     * @throws moodle_exception
+     */
+    public function __construct($manipulator, stdClass $config) {
```
kenneth-hendricks: I don't see what we gain by instantiating this class, i would think about just making the finder method static + removing the constructor

classes/local/object_manipulator/candidates/candidates_finder.php
```diff
@@ -0,0 +1,68 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Class candidates_factory
```
kenneth-hendricks: Doc class is wrong

classes/local/object_manipulator/candidates/candidates_finder.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->factory = new candidates_factory($manipulator, $config);
+    }
+
+    /**
+     * @return array
+     * @throws moodle_exception
+     * @throws dml_exception
+     */
+    public function get() {
+        $this->finder = $this->factory->finder();
```
kenneth-hendricks: I would set the finder in the constructor, if get_query_name is called before get we'll get a syntax error, currently we have an implicit ordering of how this must be called.

classes/local/object_manipulator/candidates/candidates_finder.php
```diff
@@ -0,0 +1,68 @@
...
+        return $this->finder->get();
+    }
+
+    /**
+     * @inheritDoc
+     * @return string
+     */
+    public function get_query_name() {
```
kenneth-hendricks: Actually is get_query_name actually used ? i can't see anywhere that calls this method

classes/local/object_manipulator/candidates/deleter_candidates.php
```diff
@@ -0,0 +1,65 @@
...
+                       o.location';
+    }
+
+    /**
+     * @inheritDoc
+     * @return array
+     */
+    public function get_candidates_sql_params() {
+        $consistancythrehold = time() - $this->config->consistencydelay;
+        return [$consistancythrehold, OBJECT_LOCATION_DUPLICATED, $this->config->sizethreshold];
```
kenneth-hendricks: i would consider switching to named params, i often find it them easier to read. See https://docs.moodle.org/dev/Data_manipulation_API#Placeholders

classes/local/object_manipulator/candidates/manipulator_candidates_base.php
```diff
@@ -0,0 +1,58 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Class candidates_factory
```
kenneth-hendricks: Wrong doc class

classes/local/object_manipulator/candidates/deleter_candidates.php
kenneth-hendricks: This applies to all the queries, i wont repeat myself

classes/local/object_manipulator/candidates/pusher_candidates.php
```diff
@@ -0,0 +1,91 @@
...
+class pusher_candidates extends manipulator_candidates_base {
+
+    /**
+     * pusher_candidates constructor.
+     * @param stdClass $config
+     */
+    public function __construct(stdClass $config) {
+        parent::__construct($config);
+        $this->config->maximumfilesize = \tool_objectfs\tests\test_client::MAX_UPLOAD;
+        if (in_array($config->filesystem, ['tool_objectfs\digitalocean_file_system', 'tool_objectfs\s3_file_system'])) {
```
kenneth-hendricks: So this is definitely a code smell to have a big if else chain like this when we're using a class system, can we refactor this to just call a static method on the client class instead

classes/local/object_manipulator/candidates/pusher_candidates.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @inheritDoc
+     * @return array
+     */
+    public function get_candidates_sql_params() {
+        return [
+            'maxcreatedtimstamp' => time() - $this->config->minimumage,
+            'threshold' => $this->config->sizethreshold,
+            'maximum_file_size' => $this->config->maximumfilesize, // TODO fetch by client.
```
kenneth-hendricks: leftover TODO ?

classes/local/object_manipulator/manipulator_builder.php
```diff
@@ -0,0 +1,112 @@
...
+use coding_exception;
+use dml_exception;
+use moodle_exception;
+use stdClass;
+use tool_objectfs\local\object_manipulator\candidates\candidates_finder;
+use tool_objectfs\log\aggregate_logger;
+
+defined('MOODLE_INTERNAL') || die();
+
+class manipulator_builder {
```
kenneth-hendricks: I Feel like this class is trying to do two different things  - both build the manipulators and execute them, i'd rather one of two things happen

1. We have a separate build class that builds the objects and they are executed by another
2. Only expose the execute and execute_all methods as public and have the building be implicit within them and not called separately.

classes/task/check_objects_location.php
```diff
@@ -44,8 +47,6 @@ public function get_name() {
      * Execute task
      */
     public function execute() {
-        manipulator::setup_and_run_object_manipulator('\\tool_objectfs\\local\\object_manipulator\\checker');
+        (new manipulator_builder())->build(checker::class)->execute();
```
kenneth-hendricks: This line is repeated in 5 tasks can we abstract this away so they only need to set their associated class.

classes/local/object_manipulator/candidates/candidates_finder.php
gleimermora: updated

classes/local/object_manipulator/candidates/candidates_finder.php
gleimermora: I forgot to use to call it. It is used to log the name when fetching the candidates

classes/local/object_manipulator/candidates/deleter_candidates.php
gleimermora: Thanks, I added the missing placeholders.

classes/local/object_manipulator/candidates/manipulator_candidates_base.php
gleimermora: updated

classes/local/object_manipulator/candidates/pusher_candidates.php
gleimermora: removed

classes/local/object_manipulator/manipulator_builder.php
gleimermora: I chose the option 2 :smile:

classes/local/object_manipulator/candidates/candidates_finder.php
gleimermora: good idea, thanks :+1:

classes/local/object_manipulator/candidates/candidates_factory.php
gleimermora: You are right, in this case it seems to be a good idea, but personally I'm not a friend of static methods, I try to avoid them if possible. I've update it to use `static method`.

classes/task/check_objects_location.php
gleimermora: I added the abstract class  :)

classes/local/object_manipulator/candidates/pusher_candidates.php
gleimermora: updated

kenneth-hendricks: So changing this to a map doesn't really solve this in any way - the class pusher_candidates shouldn't have any knowledge at all about the different client flavours.

As we already have the filesystem in manipulator_builder->execute, i would call $filesystem->get_external_client and then pass the filesystem client into all the manipulators. That way any of the manipulators can delegate client specific details to the client itself. This should also avoid any issues you're encountering due to 

`MicrosoftAzure\Storage\Common\Internal\Resources`

classes/local/object_manipulator/candidates/deleter_candidates.php
```diff
@@ -0,0 +1,64 @@
...
+     * @inheritDoc
+     * @return string
+     */
+    public function get_candidates_sql() {
+        return 'SELECT MAX(f.id),
+                       f.contenthash,
+                       MAX(f.filesize) AS filesize
+                  FROM {files} f
+                  JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
+                 WHERE o.timeduplicated <= :consistancythrehold
```
kenneth-hendricks: consistancythrehold instead of consistancythreshold

gleimermora: updated

PR #241, opened 2020-03-16
Add admin settings for CloudFrontClient

gleimermora:


classes/local/manager.php
```diff
@@ -29,6 +29,7 @@ public static function set_objectfs_config($config) {
     }
 
     public static function get_objectfs_config() {
+        global $CFG;
```
brendanheywood: missing php doc above? why didn't ci fail this?

gleimermora: updated

classes/local/store/s3/client.php
```diff
@@ -390,6 +395,19 @@ public function support_presigned_urls() {
...
+        if ($contentdisposition !== '') {
+            $params['ResponseContentDisposition'] = $contentdisposition;
+        }
+
+        $contenttype = $this->get_header($headers, 'Content-Type');
+        if ($contenttype !== '') {
+            $params['ResponseContentType'] = $contenttype;
+        }
+
+        if ($this->signingmethod == 'cf') {
```
brendanheywood: This method is a bit mixed with the old s3 and the new cf stuff, can you please move all the old s3 stuff to another method eg like

generate_presigned_url_cloudfront
generate_presigned_url_s3

and this method just swaps between them

classes/local/store/s3/client.php
```diff
@@ -390,6 +395,26 @@ public function support_presigned_urls() {
      * @return string.
      */
     public function generate_presigned_url($contenthash, $headers) {
+        $client = 'cf' === $this->signingmethod ? 'cloudfront' : 's3';
+        return  $this->{"generate_presigned_url_$client"}($contenthash, $headers);
```
brendanheywood: I generally dislike code like this because it makes it harder to find stuff. If you search for generate_presigned_url_s3 then you won't find where it is called from

settings.php
```diff
@@ -89,19 +87,22 @@
...
+    $config = \tool_objectfs\local\manager::get_objectfs_config();
     $support = false;
     if (!empty($config->filesystem)) {
         $support = (new $config->filesystem())->supports_presigned_urls();
     }
     $warning = !method_exists('file_system', 'supports_xsendfile');
-    $coresupport = $warning ? $OUTPUT->notification(get_string('settings:presignedurl:coresupport', 'tool_objectfs')) : '';
+    $warningtext = $warning ? $OUTPUT->notification(get_string('settings:presignedurl:coresupport', 'tool_objectfs')) : '';
+    if ( '' === $warningtext) {
+        $warningtext = \tool_objectfs\local\manager::cloudfront_pem_exists();
```
brendanheywood: should this be appending to the warning text?

Issue #243, opened 2020-03-18
Refactor the tool_objectfs_filesystem_supports_presigned_urls into each store


brendanheywood:
This function be removed:

https://github.com/catalyst/moodle-tool_objectfs/blob/master/lib.php#L179-L186

And be replace by a method in each store eg function supports_presigned_urls()

The base class should return false, and then just the s3 store should return true

PR #244, opened 2020-03-18
Add supports_presigned_urls to the store file system #243

gleimermora:
Issue #243

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -706,6 +706,15 @@ abstract class object_file_system extends \file_system_filedir {
         return false;
     }
 
+
+    /**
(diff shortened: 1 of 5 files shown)


classes/local/store/azure/file_system.php
```diff
@@ -37,4 +37,11 @@ protected function initialise_external_client($config) {
         $asclient = new client($config);
         return $asclient;
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function supports_presigned_urls() {
+        return false;
```
brendanheywood: do we need this? why not just inherit from parent?

gleimermora: updated

Issue #245, opened 2020-03-18
Refactor everything non essential out of lib.php


brendanheywood:
Should be in an autoloaded class, or classes

the _cron function must stay there for very old legacy code

PR #251, opened 2020-03-24
Clean up lib.php, move functions to static methods in config helper class

gleimermora:
…lass #245


classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+        $client = str_replace('tool_objectfs\\', 'tool_objectfs\\local\\store\\', $client.'\\client');
+
+        if (class_exists($client)) {
+            return new $client($config);
+        }
+
+        return false;
+    }
+
+    public static function tool_objectfs_get_fs_list() {
```
brendanheywood: the prefix tool_objectfs_ is not needed

classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+        $found[''] = 'Please, select';
+        $found['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system';
+        $found['\tool_objectfs\digitalocean_file_system'] = '\tool_objectfs\digitalocean_file_system';
+        $found['\tool_objectfs\s3_file_system'] = '\tool_objectfs\s3_file_system';
+        $found['\tool_objectfs\swift_file_system'] = '\tool_objectfs\swift_file_system';
+        return $found;
+    }
+
+
+    public static function update_object_by_hash($contenthash, $newlocation) {
```
brendanheywood: ditto tool_objectfs_

classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+
+class config_helper {
+
+    public static function set_objectfs_config($config) {
+        foreach ($config as $key => $value) {
+            set_config($key, $value, 'tool_objectfs');
+        }
+    }
+
+    public static function get_objectfs_config() {
```
brendanheywood: this could probably be just get_config

classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_objectfs;
```
brendanheywood: This should be inside the local namespace

classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+define('OBJECT_LOCATION_DUPLICATED', 1);
+define('OBJECT_LOCATION_EXTERNAL', 2);
+
+define('OBJECTFS_REPORT_OBJECT_LOCATION', 0);
+define('OBJECTFS_REPORT_LOG_SIZE', 1);
+define('OBJECTFS_REPORT_MIME_TYPE', 2);
+
+define('OBJECTFS_BYTES_IN_TERABYTE', 1099511627776);
+
+class config_helper {
```
brendanheywood: give that this class contains a few things which are unrelated to config, how about calling this something more generic like ' tool_objectfs\local\manager' ?

classes/config_helper.php
```diff
@@ -0,0 +1,154 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_objectfs;
+
+use stdClass;
+
+defined('MOODLE_INTERNAL') || die();
```
brendanheywood: These defines may not work correctly if this new class hasn't been autoloaded before you attempt to use them.

I think we should refactor them too, but I think lets keep that out of scope for this change. Later when we tackle them they are probably better done as class constants not defines

classes/local/object_manipulator/manipulator.php
```diff
@@ -27,6 +27,7 @@
 
 use dml_exception;
 use stdClass;
+use tool_objectfs\local\manger as config;
```
brendanheywood: Why are you aliasing this as config?

config::update_object doesnt make much sense vs

manager::update_object

brendanheywood:
good to merge when green

PR #253, opened 2020-03-30
Add private key as a textarea admin setting instead of path #252

gleimermora:


classes/local/store/s3/client.php
```diff
@@ -472,19 +462,30 @@ private function generate_presigned_url_cloudfront($contenthash, $headers = [],
...
-        $signedurl = (string)$signedurlcannedpolicy;
-
-        /* $headers[] = 'Location:"' . $signedurl . '"'; // This may cause loss of headers (etag for example). */
-        return $signedurl;
+        $json = '{"Statement":[{"Resource":"' . $resource . '","Condition":{"DateLessThan":{"AWS:EpochTime":' . $expires . '}}}]}';
+
+        // Create the private key.
+        $key = openssl_get_privatekey($this->config->cloudfrontprivatekey);
+        if (!$key) {
+            throw new \Exception('Loading private key failed');
```
brendanheywood: would this be better as a moodle_exception? and below

settings.php
```diff
@@ -97,7 +97,6 @@
     }
     $warning = !method_exists('file_system', 'supports_xsendfile');
     $warningtext = $warning ? $OUTPUT->notification(get_string('settings:presignedurl:coresupport', 'tool_objectfs')) : '';
-    $warningtext .= \tool_objectfs\local\manager::cloudfront_pem_exists();
```
brendanheywood: Can you still keep this warning in place, in case the pem config item is garbage etc

classes/local/store/s3/client.php
```diff
@@ -472,19 +462,30 @@ private function generate_presigned_url_cloudfront($contenthash, $headers = [],
...
-        ];
-        $signedurlcannedpolicy = $client->getSignedUrl($signingparameters);
-        $signedurl = (string)$signedurlcannedpolicy;
-
-        /* $headers[] = 'Location:"' . $signedurl . '"'; // This may cause loss of headers (etag for example). */
-        return $signedurl;
+        $json = '{"Statement":[{"Resource":"' . $resource . '","Condition":{"DateLessThan":{"AWS:EpochTime":' . $expires . '}}}]}';
+
+        // Create the private key.
+        $key = openssl_get_privatekey($this->config->cloudfrontprivatekey);
```
brendanheywood: Just being my own devils advocate: in the auth_saml plugin we can accept certs either inline or as a url, we probably want similar flexibility here too, ie in some cases like this dev env we want to get up and running quickly inline but in prod we probably want to lock it down using a path which is forced. So I'm thinking of best of both worlds where we can accept either an inline cert or a path to a cert.

Should only be a couple lines, if the config item starts with '/' and it's a single line then treat it as a file path, if not treat as inline cert

classes/local/store/s3/client.php
brendanheywood: also 'Loading private key failed' doesn't mention what plugin or config we are dealing with, maybe something like 'tool_object could not load cloudfront signing key'

classes/local/store/s3/client.php
```diff
@@ -472,19 +462,30 @@ private function generate_presigned_url_cloudfront($contenthash, $headers = [],
...
+        if (!openssl_sign($json, $signedpolicy, $key, OPENSSL_ALGO_SHA1)) {
+            throw new \Exception('Signing policy failed, ' . openssl_error_string());
+        }
+
+        // Create url safe signed policy.
+        $base64signedpolicy = base64_encode($signedpolicy);
+        $signature = str_replace(['+', '=', '/'], ['-', '_', '~'], $base64signedpolicy);
+
+        // Construct the URL.
+        return $resource . (strpos($resource, '?') === false ? '?' : '&') . 'Expires=' . $expires .
```
brendanheywood: this might be cleaner using moodle_url($resource, $overrideparams)

settings.php
brendanheywood: the alternative would be a custom admin setting for the pem with it's own validation. I'll make that as a new issue

brendanheywood: https://github.com/catalyst/moodle-tool_objectfs/issues/254

classes/local/store/s3/client.php
gleimermora: updated

classes/local/store/s3/client.php
gleimermora: I'll implement in #254

Issue #254, opened 2020-03-30
Make admin setting pem a custom admin setting


brendanheywood:
the alternative would be a custom admin setting for the pem with it's own validation. I'll make that as a new issue

Validates:
* [x] looks like a path, and it points to a file which exists and is readable
* [x] or looks like a pem
* [x] either way it can actually be opened, parsed by openssl

PR #255, opened 2020-03-31
Cloudfront private key setting file or key content #254

gleimermora:


classes/local/manager.php
```diff
@@ -196,20 +195,31 @@ static public function cloudfront_pem_exists() {
...
+    /**
+     * @param string $cloudfrontprivatekey
+     * @return bool|resource
+     */
+    static public function parse_cloudfront_private_key($cloudfrontprivatekey) {
+        global $CFG;
+        $path = $CFG->dataroot . $cloudfrontprivatekey;
+        if (file_exists($path) && is_readable($path)) {
+            $cert = file_get_contents($path);
+            $parsedkey = openssl_pkey_get_private($cert);
```
brendanheywood: why not close the if here and have the same code path after this?

lang/en/tool_objectfs.php
```diff
@@ -162,11 +162,11 @@
 $string['settings:presignedcloudfronturl:cloudfront_key_pair_id'] = 'Key_Pair ID from AWS';
 $string['settings:presignedcloudfronturl:cloudfront_key_pair_id_help'] = 'This is generated using AWS account \'root\' user (along with the private key .pem file).';
 $string['settings:presignedcloudfronturl:cloudfront_private_key_pem'] = 'PRIVATE Key .pem';
-$string['settings:presignedcloudfronturl:cloudfront_private_key_pem_help'] = 'Private key content (.pem file).';
+$string['settings:presignedcloudfronturl:cloudfront_private_key_pem_help'] = 'Private key content (.pem file) or relative path to cloudfront.pem file. ($CFG->dataroot . "/cloudfront.pem")';
```
brendanheywood: This string seems inconsistent with the implementation?

classes/local/manager.php
gleimermora: done

lang/en/tool_objectfs.php
gleimermora: updated

PR #258, opened 2020-04-01
Add tests/fixture file to test relative URLs

gleimermora:


settings.php
```diff
@@ -114,6 +114,24 @@
...
+        $options = [
+            'video/mp4' => 'video/mp4',
+            'text/plain' => 'text/plain',
+            'image/png' => 'image/png',
+            'text/html' => 'text/html',
+            'text/css' => 'text/css',
+            'application/pdf' => 'application/pdf'
+        ];
+        $settings->add(
+            new admin_setting_configmulticheckbox(
```
brendanheywood, line 126: Is it possible to use admin_setting_filetypes ?

tests/fixtures/testrelativeurls.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>HTML testing relative URLs</title>
```
brendanheywood, line 4: can you add a test with css as well please?

classes/local/store/object_file_system.php
```diff
@@ -751,11 +756,62 @@ public function presigned_url_should_redirect($contenthash) {
      * @param array $headers request headers.
      *
      * @return string.
+     * @throws \dml_exception
      */
     public function generate_presigned_url_to_external_file($contenthash, $headers = array()) {
+        if ($this->is_white_listed($contenthash)) {
+            return $this->get_url_by_contenthash($contenthash);
```
brendanheywood, line 763: isn't this dead code? if the file type is black listed then we should not get here, if we are white listed then the signed url shouldn't change?

classes/local/store/object_file_system.php
```diff
@@ -736,6 +736,11 @@ public function presigned_url_should_redirect($contenthash) {
             return true;
         }
 
+        // If file MIME type is whitelisted we do not generate a presigned_url for it.
```
brendanheywood, line 739: I think we are doing more db calls than we need to, for this logic and the logic below around file size

Is there any way to infer the file type from the current http request without doing this db lookup? If not can we do 1 db lookup to grab the record and then use that for both bits of logic below?

peterspicer-catalyst: I ran into the same frustration, I ended up trying to do it later in the process once the Content-Type that was to be sent to the user was known and inferred it from that, then threw an exception which bubbled up and could be caught to safely abort the redirect.

This worked fine for the pre-signed test page but not for SCORM, possibly because I think (but didn't get around to fixing) that by the time I was doing it, Content-Type was mangled to include a suffix for indicating UTF-8 if it was otherwise text/html. But I ran out of time I could spend on it :(

settings.php
gleimermora: updated

renderer.php
```diff
@@ -280,7 +280,7 @@ public function presignedurl_tests_content($fs, $testfiles) {
         $output .= $this->heading(get_string('presignedurl_testing:test1', 'tool_objectfs'), 4);
         foreach ($testfiles as $file) {
             $headers = array('Content-Disposition: attachment');
-            $presignedurl = $fs->generate_presigned_url_to_external_file($file->get_contenthash(), $headers);
+            $presignedurl = $this->generate_presigned_url($fs, $file, $headers);
```
brendanheywood, line 283: $CFG->enablepresignedurls = true;

brendanheywood: 2) swap to a true whitelist

3) add a test file reset button

classes/local/store/object_file_system.php
```diff
@@ -756,6 +762,29 @@ public function generate_presigned_url_to_external_file($contenthash, $headers =
...
+     * @param string $filename
+     * @return bool
+     * @throws \dml_exception
+     */
+    public function is_extension_whitelisted($filename) {
+        $config = manager::get_objectfs_config();
+        if (empty($config->signingwhitelist)) {
+            return false;
+        }
+        $whitelist = explode(',',  $config->signingwhitelist);
```
brendanheywood, line 776: just be careful with whitespace, this can be edited by humans directly and whitespace around each item should be trimmed

classes/local/store/object_file_system.php
```diff
@@ -756,6 +762,29 @@ public function generate_presigned_url_to_external_file($contenthash, $headers =
...
+        if (empty($whitelist)) {
+            return false;
+        }
+        $extension = pathinfo($filename, PATHINFO_EXTENSION);
```
brendanheywood, line 780: It would be better to not implement this logic, there should be a core helper for this.

Look at other code which uses the admin_setting_filetypes, also talk to peter as he did work on this in another plugin which is very similar:

https://tracker.moodle.org/browse/MDL-54716

https://github.com/moodle/moodle/compare/master...Peterburnett:MDL-54716#diff-f2b5c8f463bfa8a6613933d9b305cf57R784-R785

     $util = new core_form\filetypes_util();
        $sitewidetypes = $util->normalize_file_types($CFG->allowedfiletypes);

lang/en/tool_objectfs.php
```diff
@@ -153,6 +153,9 @@
 $string['settings:presignedurl:enablepresigneds3urls'] = 'S3 Pre-Signed URLs';
 $string['settings:presignedurl:enablepresigneds3urls_help'] = 'Enable Pre-Signed S3 URLs to request content directly from external storage.';
 
+$string['settings:presignedurl:whitelist'] = 'Pre-Signed URL whitelist.';
+$string['settings:presignedurl:whitelist_help'] = 'Files with the selected MIME types will generate a local url.';
```
brendanheywood, line 157: should this say extension instead of mimetype?

brendanheywood: and signed url, instead of local url

renderer.php
```diff
@@ -328,4 +328,27 @@ public function presignedurl_tests_content($fs, $testfiles) {
...
+     * @param $fs
+     * @param \stored_file $file
+     * @param array $headers
+     * @return string
+     * @throws dml_exception
+     */
+    public function generate_presigned_url($fs, $file, array $headers = []) {
+        $filename = $file->get_filename();
+        if ($fs->is_extension_whitelisted($filename)) {
+            return \moodle_url::make_pluginfile_url(
```
brendanheywood, line 343: This feels wrong to be here. I suspect the only reason you need this is because of the testing page, but normal files shouldn't need this. For normal files you should not have gotten this far because the white list would have kicked in earlier.

The problem is that different plugins call moodle_url::make_pluginfile_url in different ways, so this code path while it may work for the test page, isn't 100% correct if called from other paths of moodle. I'd prefer it to be safe and simply return false or null or throw an exception, and it's up to the calling page to default back to the original path.

lang/en/tool_objectfs.php
gleimermora: updated

renderer.php
gleimermora: this is used only for the test page, why would it be called from somewhere else?

presignedurl_tests.php
```diff
@@ -35,18 +35,26 @@
...
     $support = $fs->supports_presigned_urls();
 }
-$settingslink = \html_writer::link(new \moodle_url('/admin/settings.php?section=tool_objectfs'),
-    get_string('presignedurl_testing:objectfssettings', 'tool_objectfs'));
-
 if ($support) {
-
+    if ($delete === 1) {
+        $output->delete_presignedurl_tests_files();
+        echo $output->notification('Files deteled successfully', 'success');
```
brendanheywood, line 51: typo with 'deteled', also please move this to a lang string

presignedurl_tests.php
```diff
@@ -35,18 +35,26 @@
...
     $fs = new $config->filesystem();
     $support = $fs->supports_presigned_urls();
 }
-$settingslink = \html_writer::link(new \moodle_url('/admin/settings.php?section=tool_objectfs'),
-    get_string('presignedurl_testing:objectfssettings', 'tool_objectfs'));
-
 if ($support) {
-
+    if ($delete === 1) {
+        $output->delete_presignedurl_tests_files();
```
brendanheywood, line 50: You don't check the sesskey here. It's not critical in this case but it's a good habit to keep up

presignedurl_tests.php
```diff
@@ -35,18 +35,26 @@
...
+        echo $output->notification('Files deteled successfully', 'success');
+        $url = new \moodle_url('/admin/tool/objectfs/presignedurl_tests.php');
+        echo $output->heading(html_writer::link($url, 'Reload page'), 6);
+        echo $output->footer();
```
brendanheywood, line 54: If you render the footer you should probably just exit after that.

But in this case why not just move the logic further up so you don't need a reload after the delete?

renderer.php
```diff
@@ -232,6 +232,31 @@ public function object_status_page_intro() {
...
+    /**
+     * Delete test files from files table
+     * @throws dml_exception
+     */
+    public function delete_presignedurl_tests_files() {
+        global $CFG;
+        $filestorage = get_file_storage();
+        $fixturespath = $CFG->dirroot . '/admin/tool/objectfs/tests/fixtures/';
+        $fixturesfiles = glob($fixturespath.'*');
+        foreach ($fixturesfiles as $fixturesfile) {
```
brendanheywood, line 244: This should not delete all the fixtures, it should delete all the files in the file api file area

brendanheywood: https://docs.moodle.org/dev/File_API#List_area_files

eg

```
$fs = get_file_storage();
$files = $fs->get_area_files($contextid, 'mod_assignment', 'submission', $submission->id);
foreach ($files as $f) {
    $f->delete();
}
```

renderer.php
```diff
@@ -274,58 +299,93 @@ public function presignedurl_tests_load_files($fs) {
...
+     */
+    private function get_output($url, $filename, $identifier) {
+        if ('iframesnotsupported' === $identifier) {
+            $output = '<iframe height="400" width="100%" src="' . $url . '">'.
+                get_string('presignedurl_testing:' . $identifier, 'tool_objectfs').'</iframe>';
+        } else {
+            $output = get_string('presignedurl_testing:' . $identifier, 'tool_objectfs').': '.
+                '<a href="'. $url .'">'. $filename . '</a>';
+        }
+        return $output . '<br><small>Redirecting to: ' . pathinfo($url, PATHINFO_DIRNAME) . '/...</small>';;
```
brendanheywood, line 366: I like this, can you improve this so that it shows when it will or won't redirect?

renderer.php
```diff
@@ -274,58 +299,93 @@ public function presignedurl_tests_load_files($fs) {
...
+     * Generates the output string that contains the presignedurl.
+     * @param string $url
+     * @param string $filename
+     * @param string $identifier
+     * @return string
+     * @throws coding_exception
+     */
+    private function get_output($url, $filename, $identifier) {
+        if ('iframesnotsupported' === $identifier) {
+            $output = '<iframe height="400" width="100%" src="' . $url . '">'.
```
brendanheywood, line 360: please always use the html_writer api so the url is escaped correctly, and below with the a tag

renderer.php
brendanheywood: You've subtly changed the public contract. If the url cannot be signed, then in other code it throws an exception and you should probably do the same here. Do we do not every want code to be able to call this and get a url back when it shouldn't. If it did you are just burying a bug that is going to bite someone in the future.

settings.php
```diff
@@ -114,6 +114,16 @@
             new lang_string('settings:presignedurl:presignedminfilesize', 'tool_objectfs'),
             new lang_string('settings:presignedurl:presignedminfilesize_help', 'tool_objectfs'), 0, PARAM_INT));
 
+        $settings->add(
+            new admin_setting_filetypes(
+                'tool_objectfs/signingwhitelist',
+                new lang_string('settings:presignedurl:whitelist', OBJECTFS_PLUGIN_NAME),
+                new lang_string('settings:presignedurl:whitelist_help', OBJECTFS_PLUGIN_NAME),
+                '',
+                ['onlytypes' => ['txt', 'image', 'document', 'web_file', 'video']]
```
brendanheywood, line 123: this might be the source of the bug where you can't save the admin setting when you tick lots of things

classes/local/store/object_file_system.php
```diff
@@ -756,6 +762,30 @@ public function generate_presigned_url_to_external_file($contenthash, $headers =
         return $this->externalclient->generate_presigned_url($contenthash, $headers);
     }
 
+    /**
+     * Check if file extension is whitelisted.
+     * @param string $filename
+     * @return bool
+     * @throws \dml_exception
+     */
+    public function is_extension_whitelisted($filename) {
```
brendanheywood, line 771: this seems to be reinventing the wheel, why can't you just call:

```
$util = new filetypes_util();
$util->is_whitelisted('txt', 'doc txt rtf')
```

renderer.php
gleimermora: the contract is still the same, this is only used for the test page it isn't used.
This is the public method that is used anywhere else: https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/object_file_system.php#L755 so the current behaviour remains the same.

classes/local/store/object_file_system.php
gleimermora: updated

renderer.php
gleimermora: this url has already been properly generated.

presignedurl_tests.php
```diff
@@ -33,20 +33,30 @@
 
 $output = $PAGE->get_renderer('tool_objectfs');
 
+$delete = optional_param('delete', 0, PARAM_INT);
```
brendanheywood, line 36: This would be touch nicer as a PARAM_BOOL and then you don't need to === below

Issue #270, opened 2020-04-09
Presigned URLs: check if expirationtime equals to Thu, 01 Jan 1970 00:00:00 GMT


golenkovm:
https://github.com/catalyst/moodle-tool_objectfs/blob/11b687c8bcaaa32f4c4b976aec2ac4fb2f611f79/classes/local/store/s3/client.php#L437

This should treat `$expirationtime == ' Thu, 01 Jan 1970 00:00:00 GMT'` to be the same as empty.
Otherwise I had an error with S3 presigned urls:

```<Error>
<Code>AuthorizationQueryParametersError</Code>
<Message>X-Amz-Expires must be non-negative</Message>
<RequestId>BEF94637B4AACC66</RequestId>
<HostId>
UEI4T2AgnrAiZ3muQq2Vh2Di5jo/hSfRwYlUm2VoaFjJPsDtzBKLeIE3b7eluyoMS+OK4ksMRTo=
</HostId>
</Error>```

golenkovm:
This has been fixed by merging #293 into master branch.
Closing this off.

PR #275, opened 2020-04-13
Validate  is in seconds #270

gleimermora:
#270


renderer.php
```diff
@@ -306,94 +307,106 @@ public function presignedurl_tests_content($fs, $testfiles) {
...
 
+    /**
+     * Generate a file url with adding a param to set 'Expires' header.
+     * @param stored_file $file
+     * @param int|bool $expires
+     * @param bool $forcedownload
+     * @return string
+     * @throws dml_exception
+     */
+    private function generate_presigned_url($file, $expires = false, $forcedownload = false) {
```
brendanheywood, line 372: I think the name of this is misleading, this url is not presigned

Issue #279, opened 2020-04-16
Minimum size for Pre-Signed URL should not ignore file extension check


gleimermora:
Currently, objectfs redirects all urls when `presignedminfilesize` is set to `0`. 
It should check file extension as well.

golenkovm:
This has been fixed

PR #280, opened 2020-04-16
Re-work presigned_url_should_redirect() method

gleimermora:
This is a pull request fot #279 

1. Implement new manager method `all_extensions_whitelisted()`.
2. Always redirect when `presignedminfilesize` is set to `0` and all file extentions are whitelisted.
3. Otherwise, redirect when the file size is bigger than presignedminfilesize setting and the file extension is whitelisted.
4. Fix `test_presigned_url_should_redirect_provider()` unit test and set `signingwhitelist` to `*`, as all files created in this test have hash as their names.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -737,22 +737,22 @@ abstract class object_file_system extends \file_system_filedir {
         }
 
         // Redirect regardless.
-        if ($this->externalclient->presignedminfilesize == 0) {
+        if ($this->externalclient->presignedminfilesize == 0 &&
+                manager::all_extensions_whitelisted()) {
             return true;
         }
 
-        // Redirect only files that bigger than configured value.
-        // And if file extension is not whitelisted.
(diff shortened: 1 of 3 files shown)


classes/local/store/object_file_system.php
```diff
@@ -731,11 +731,6 @@ public function presigned_url_should_redirect($contenthash) {
...
         }
 
-        // Redirect regardless.
-        if ($this->externalclient->presignedminfilesize == 0) {
-            return true;
-        }
-
         // Redirect only files that bigger than configured value.
         // And if file extension is not whitelisted.
         if ($this->externalclient->presignedminfilesize > 0) {
```
brendanheywood, line 736: is this if still valid when presignedminfilesize == 0 ??

There should be a fallback return at the end of this function, or the if should be removed

gleimermora: yes, the if has be go

brendanheywood:
this is good to merge when green

Issue #282, opened 2020-04-20
admin_setting_filetypes doesn't exist in some legacy/totara installations


Peterburnett:
Settings page is inaccessable. It may be useful to define an alternate admin setting to use, such as admin_setting_configtextarea, that can be used on legacy. They should be read almost identically by core, as they both should mostly boil down to a comma seperated list.

brendanheywood:
I'd aim to poly fill this, ie detect if the better one is present and if not just swap to config text area

More of a worry is all the core libs behind it which are probably missing too.

brendanheywood:
More context:

```
Exception - Class 'admin_setting_filetypes' not found
More information about this error
Debug info:
Error code: generalexceptionmessage
Stack trace:

    line 125 of /admin/tool/objectfs/settings.php: Error thrown
    line 76 of /lib/classes/plugininfo/tool.php: call to include()
    line 557 of /admin/settings/plugins.php: call to core\plugininfo\tool->load_settings()
    line 8001 of /lib/adminlib.php: call to require()
    line 443 of /lib/classes/plugininfo/base.php: call to admin_get_root()
    line 1254 of /admin/renderer.php: call to core\plugininfo\base->get_settings_url()
    line 343 of /admin/renderer.php: call to core_admin_renderer->plugins_control_panel()
    line 166 of /admin/plugins.php: call to core_admin_renderer->plugin_management_page()
```

this was on totara:

```
$version  = 2016120509.00;              // YYYYMMDD      = weekly release date of this DEV branch.
                                        //         RR    = release increments - 00 in DEV branches.
                                        //           .XX = incremental changes.

$release  = '3.2.9 (Build: 20180517)'; // Human-friendly version name

$branch   = '32';                       // This version's branch.

...

$TOTARA->version    = '11.25';          // Please keep as string.
$TOTARA->build      = '20200327.00';   // Please keep as string.
```

As we cannot unit test totara here lets see if that bug also exists in the same way on moodle 3.2 (or 3.1 LTS which we are already testing in travis)

golenkovm:
Closing this issue as it was fixed by merging #283 into master branch.
I've created a separate issue #291 to backport necessary stuff like `admin_setting_filetypes` and let PreSigned URLs be configured in legacy/totara installations

PR #283, opened 2020-04-21
Check that admin_setting_filetypes class exists

golenkovm:
Fix for #282

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -104,75 +104,83 @@ if ($ADMIN->fulltree) {
         }
     }
 
+    $warningtext = '';
     $signingsupport = false;
     if (!empty($config->filesystem)) {
         $signingsupport = (new $config->filesystem())->supports_presigned_urls();
     }
-    $warning = !method_exists('file_system', 'supports_xsendfile');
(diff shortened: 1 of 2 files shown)


settings.php
```diff
@@ -104,75 +104,83 @@
...
         $signingsupport = (new $config->filesystem())->supports_presigned_urls();
     }
-    $warning = !method_exists('file_system', 'supports_xsendfile');
-    $warningtext = $warning ? $OUTPUT->notification(get_string('settings:presignedurl:coresupport', 'tool_objectfs')) : '';
+    if (!method_exists('file_system', 'supports_xsendfile')) {
+        $warningtext .= $OUTPUT->notification(get_string('settings:presignedurl:coresupport', 'tool_objectfs'));
+    }
     $warningtext .= \tool_objectfs\local\manager::cloudfront_pem_exists();
+    $classexists = class_exists('admin_setting_filetypes');
+    if (!class_exists('admin_setting_filetypes')) {
```
brendanheywood, line 117: These two lines are the same, did you meant o use $classexists?

golenkovm: Ah, yes, initially it was `if (!$classexists) {`

golenkovm: Fixed

Issue #284, opened 2020-04-21
Setting filesystem to Azure without having SDK libraries installed breaks admin page


golenkovm:
After setting to `\tool_objectfs\azure_file_system`:
![image](https://user-images.githubusercontent.com/22066290/79833492-cba0f580-83ee-11ea-9921-daca76805377.png)

golenkovm:
Fixed by merging #285 into master branch

PR #285, opened 2020-04-21
Avoid errors when storage SDK is not installed

golenkovm:
This is the fix for #284

### DIFF
--- a/classes/local/manager.php
+++ b/classes/local/manager.php
@@ -111,29 +111,14 @@ class manager {
      * @return bool
      */
     public static function get_client($config) {
-        $fsclass = $config->filesystem;
-        $client = str_replace('_file_system', '', $fsclass);
-        $client = str_replace('tool_objectfs\\', 'tool_objectfs\\local\\store\\', $client.'\\client');
+        $clientclass = self::get_client_classname_from_fs($config->filesystem);
 
-        if (class_exists($client)) {
-            return new $client($config);
+        if (class_exists($clientclass)) {
+            return new $clientclass($config);
         }
 
         return false;
     }
-
-    /**
-     * @return mixed
-     */
-    public static function get_fs_list() {
-        $found[''] = get_string('pleaseselect', OBJECTFS_PLUGIN_NAME);
-        $found['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system';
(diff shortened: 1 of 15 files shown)


classes/local/manager.php
```diff
@@ -134,6 +134,29 @@ public static function get_fs_list() {
...
+    /**
+     * Returns the list of installed and available filesystems.
+     *
+     * @return array
+     */
+    public static function get_available_fs_list() {
+        $result = [];
+        $filesystems = self::get_fs_list();
+        foreach ($filesystems as $filesystem) {
+            if ($filesystem == 'Please, select') {
```
dmitriim, line 146: will it work with translated strings?

dmitriim: oh, it's hard-coded errrr

golenkovm: Yep. But good catch. Can be swapped to lang string

classes/local/store/azure/client.php
```diff
@@ -27,18 +27,13 @@
...
-use MicrosoftAzure\Storage\Common\ServicesBuilder;
-use MicrosoftAzure\Storage\Common\Exceptions\ServiceException;
 use SimpleXMLElement;
 use stdClass;
 use tool_objectfs\local\store\azure\stream_wrapper;
 use tool_objectfs\local\store\object_client_base;
 
 class client extends object_client_base {
-    const MAX_UPLOAD = Resources::MAX_BLOCK_BLOB_SIZE;
+    const MAX_UPLOAD = 5242880000000;
```
brendanheywood, line 36: Can you throw a comment here saying this is should be = to  Resources::MAX_BLOCK_BLOB_SIZE

or better yet can we make this just a variable instead of a const and set it in the constructor?

classes/local/manager.php
golenkovm: This is done

classes/local/store/azure/client.php
golenkovm: Class variable has been created. I had to slightly refactor client and filesystem classes to meet this change

golenkovm: Probably, this is not needed at all: https://github.com/catalyst/moodle-tool_objectfs/pull/285/commits/d3f95c491c0767c4ce6c3a0198e4a53eb694140a#diff-18ab1659be8a675b97cef923858847adR78-R83

PR #287, opened 2020-04-22
Remove whitespace in install.xml to fix test fail

guillogo:


kristian-94:
Looks good @guillogo , could you also just add the issue number in the commit message, at the end ie. `Remove whitespace in install.xml to fix test fail #287` , that will automatically link to the issue in github and is a good habit to put into commit messages

Issue #289, opened 2020-04-22
Swift: Integration testing


golenkovm:
Running phpunit with `$CFG->phpunit_objectfs_swift_integration_test_credentials` I'm having this:

```
1) tool_objectfs\tests\checker_testcase::test_checker_get_location_local_if_object_is_local
Error: Class 'tool_objectfs\tests\test_swift_integration_client' not found

/siteroot/admin/tool/objectfs/tests/classes/test_file_system.php:71
/siteroot/admin/tool/objectfs/classes/local/store/object_file_system.php:61
/siteroot/admin/tool/objectfs/tests/tool_objectfs_testcase.php:38
/siteroot/admin/tool/objectfs/tests/checker_test.php:33
/siteroot/lib/phpunit/classes/advanced_testcase.php:80

To re-run:
 vendor/bin/phpunit -c admin/tool/objectfs/ "tool_objectfs\tests\checker_testcase" admin/tool/objectfs/tests/checker_test.php

ERRORS!
```

PR #290, opened 2020-04-22
Include Swift integration client into test_file_system class #289.

golenkovm:
Fix for #289


tuanngocnguyen:
Hi @golenkovm 

Since there are changes in the supported branches as per: https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/README.md#Branches.
Please create new MR against MOODLE_33_STABLE/MOODLE_310_STABLE if you are still working on the issue.
Thanks

Issue #294, opened 2020-04-28
Introduce unit test case for manager class


golenkovm:
Currently there is no test case for manager class

golenkovm:
The test case for manager class is added as part of #295 and can be extended to cover all manager methods, however it fails with Moodle 27 and 31 because of `Class 'core_form\filetypes_util' not found`. 
So now this issue is blocked by #292 
Proper patches for legacy versions should be implemented first.

azrek:
hi @golenkovm 

It also fails for Totara 12 - which is based on Moodle 3.3.

```
1) tool_objectfs\tests\manager_testcase::test_is_extension_whitelisted with data set #2 ('*', 'file.tar', true)
Error: Class 'core_form\filetypes_util' not found

/var/www/site/admin/tool/objectfs/classes/local/manager.php:234
/var/www/site/admin/tool/objectfs/tests/manager_test.php:98
/var/www/site/lib/phpunit/classes/base_testcase.php:600
/var/www/site/lib/phpunit/classes/advanced_testcase.php:68

To re-run:
 vendor/bin/phpunit tool_objectfs\tests\manager_testcase admin/tool/objectfs/tests/manager_test.php
```


Thanks,
Ghada

golenkovm:
Hi @azrek 
As discussed, MDL-53240 needs to be backported.
Additional exception is added in #344 and unneeded `element-filetypes.mustache` is removed in #356 
Closing this off now

PR #295, opened 2020-04-28
Introduce unit test case for manager class.

golenkovm:
This is a pull request for #294


golenkovm:
Will be fixed as part of #340

Issue #298, opened 2020-05-01
presigned_url_should_redirect SQL can return more than one record


golenkovm:
If you take an existing file and add it to Moodle with another name this sql can return more than one record:
https://github.com/catalyst/moodle-tool_objectfs/blob/4344dcd0a7f36f7c97e5ca36684c4c1d7d4518b4/classes/local/store/object_file_system.php#L747-L752

and you''ll see an error:
```
Error: mdb->get_record() found more than one record!
line 1599 of /lib/dml/moodle_database.php: call to debugging()
line 752 of /admin/tool/objectfs/classes/local/store/object_file_system.php: call to moodle_database->get_record_sql()
line 401 of /admin/tool/objectfs/renderer.php: call to tool_objectfs\local\store\object_file_system->presigned_url_should_redirect()
line 311 of /admin/tool/objectfs/renderer.php: call to tool_objectfs_renderer->get_output()
line 65 of /admin/tool/objectfs/presignedurl_tests.php: call to tool_objectfs_renderer->presignedurl_tests_content()
```

golenkovm:
Fixed by merging #299 into master branch.
Closing this off.

Issue #263, opened 2020-04-06
avoid an extra file db lookup in core


brendanheywood:
* [x] https://tracker.moodle.org/browse/MDL-68342
* [x] take advantage of the above
* [x] add the backporting details to the readme

It needs to work regardless of wether core has this backport or not

golenkovm:
Implemented and merged into master branch.
However, new php7 syntax breaks legacy moodle version with php5 so #316 was created and readme was updated.

PR #299, opened 2020-05-01
Select 1 record when checking file size extension.

golenkovm:
This is a fix for issue #298

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -697,9 +697,10 @@ abstract class object_file_system extends \file_system_filedir {
         switch ($location) {
             case OBJECT_LOCATION_EXTERNAL:
             case OBJECT_LOCATION_DUPLICATED:
-                if ($this->presigned_url_should_redirect($contenthash)) {
+                $headers = headers_list();
+                if ($this->presigned_url_should_redirect($contenthash, $headers)) {
                     try {
-                        redirect($this->generate_presigned_url_to_external_file($contenthash, headers_list()));
+                        redirect($this->generate_presigned_url_to_external_file($contenthash, $headers));
                     } catch (\Exception $e) {
                         return false;
                     }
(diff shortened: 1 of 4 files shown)


golenkovm:
#263 will make this sql redundant as file object will be passed down to objectfs.
But to fix this now I've done this quick path.

classes/local/store/object_file_system.php
```diff
@@ -748,7 +748,8 @@ public function presigned_url_should_redirect($contenthash) {
                   FROM {files}
                  WHERE contenthash = :contenthash
                    AND filesize > :filesize
-              GROUP BY filename';
+              GROUP BY filename
+                 LIMIT 1';
```
brendanheywood, line 752: Why do we need to group by filename? Why not just remove filename completely

Either that or we can just select the first record without the group by which should be even quicker.

The filesize clause also seems redundant too

golenkovm: We need file size to compare with `presignedminfilesize`. File name is needed to check if the file extension whether or not whitelisted.
But, yeah, we can just select the first row without grouping and calculating max file size.

golenkovm: Done

dmitriim: could it be different filename, but the same hash?

golenkovm: Yes, it could be. 
Same content hash means the same file type and content, so new file name is likely to have the same extension. Of course, user can change extension or remove it and this will make the difference, but as outcome the file won't be redirected to pre-signed url and will be retrieved as usual. Nothing will be broken, so we can live with this until #263 is implemented.

brendanheywood: I'm not sure we do actually need the filename in this query as we should already have access to it in possible two places:

1) the Content-Disposition (but it's optional)
2) the request path info:

I think we can split that logic out and do the whitelisting first, because if the extension isn't whitelisted we can exit early and save the db call (even through we will also save it in #263 the backwards compat performance is important too)

Issue #146, opened 2019-06-18
Dynamic list of regions


dmitriim:
New SDK allows you to have a dynamic list of regions. See https://github.com/catalyst/moodle-local_aws/issues/13

golenkovm:
This has been implemented. Closing this off.

PR #319, opened 2020-05-12
Implement dynamic list of regions.

golenkovm:
This is a fix for #146

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -29,6 +29,7 @@ defined('MOODLE_INTERNAL') || die();
 
 use tool_objectfs\local\manager;
 use tool_objectfs\local\store\object_client_base;
+use local_aws\admin_settings_aws_region;
 
 define('AWS_API_VERSION', '2006-03-01');
 define('AWS_CAN_READ_OBJECT', 0);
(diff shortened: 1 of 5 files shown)


golenkovm:
The section now looks like
image

A new `catch` clause is need to cover this exception:
image

classes/local/store/s3/client.php
```diff
@@ -29,6 +29,7 @@
 
 use tool_objectfs\local\manager;
 use tool_objectfs\local\store\object_client_base;
+use local_aws\admin_settings_aws_region;
```
brendanheywood, line 32: will this explode if not found?

lang/en/tool_objectfs.php
```diff
@@ -93,6 +93,7 @@
 $string['settings:aws:region_help'] = 'Amazon S3 API gateway region.';
 $string['settings:aws:base_url'] = 'Base URL';
 $string['settings:aws:base_url_help'] = 'Alternate url for cnames or s3 compatible endpoints. Leave blank for normal S3 use.';
+$string['settings:aws:upgradeneeded'] = 'Please upgrade \'local_aws\' plugin up to 2020051200 version.';
```
brendanheywood, line 96: you should say, "at least". Probably better to say 'to the latest supported version'

classes/local/store/s3/client.php
golenkovm: Tested this. It's not exploding.

lang/en/tool_objectfs.php
golenkovm: Set to `'Please upgrade \'local_aws\' plugin to the latest supported version.'`

Issue #261, opened 2020-04-02
Everytime we run the generate_status_report task, store a snap shot for historical trends


brendanheywood:
So we can get a daily trend of where files / space is

golenkovm:
Done. Closing this off

PR #325, opened 2020-05-14
Report: Store snapshots for historical trends

golenkovm:
This is an implementation of #261


classes/local/report/object_location_history_table.php
```diff
@@ -0,0 +1,138 @@
...
+ */
+class object_location_history_table extends \table_sql {
+
+    /**
+     * Constructor for the file location history table.
+     */
+    public function __construct() {
+        parent::__construct('locationhistory');
+
+        // TODO: Replace by lang strings.
```
brendanheywood, line 46: done!

classes/local/report/object_location_history_table.php
```diff
@@ -0,0 +1,138 @@
...
+        $sql = 'SELECT CONCAT(timecreated, datakey) AS uid,
+                       timecreated AS date,
+                       datakey AS location,
+                       objectcount AS count,
+                       objectsum AS size
+                  FROM {tool_objectfs_reports}
+                 WHERE reporttype = :reporttype
+              ORDER BY timecreated';
+        $params = array('reporttype' => 'location');
+        $rawrecords = $DB->get_records_sql($sql, $params, 0, 5000);
```
brendanheywood, line 93: this is probably better as a recordset

object_location.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Object location history page.
+ *
+ * @package   tool_objectfs
+ * @author    Mikhail Golenkov <golenkovm@gmail.com>
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require_once(__DIR__ . '/../../../config.php');
```
brendanheywood, line 26: this should declare buffering off

```
define('NO_OUTPUT_BUFFERING', true);
```

object_location.php
```diff
@@ -0,0 +1,59 @@
...
+
+$pageurl = new \moodle_url('/admin/tool/objectfs/object_location.php', $params);
+$heading = get_string('object_status:locationhistory', 'tool_objectfs');
+$PAGE->set_url($pageurl);
+$PAGE->set_context(context_system::instance());
+$PAGE->set_pagelayout('report');
+$PAGE->set_title($heading);
+$PAGE->set_heading($heading);
+
+admin_externalpage_setup('tool_objectfs_object_location_history');
```
brendanheywood, line 44: good habit to always do this right at the top first as this is the bit that enforces the security layer

object_location.php
```diff
@@ -0,0 +1,59 @@
...
+$table->baseurl = $pageurl;
+
+if (empty($logformat)) {
+    echo $OUTPUT->header();
+    $table->out(100, false);
+    echo $OUTPUT->footer();
+} else {
+    $filename = 'object_location_history_' . userdate(time(), get_string('backupnameformat', 'langconfig'), 99, false);
+    $table->is_downloading($logformat, $filename);
+    $table->out(100, false);
```
brendanheywood, line 58: you probably don't want a pagesize here, ie download the lot

classes/local/report/objectfs_report.php
```diff
@@ -97,14 +108,18 @@ public static function get_last_generate_status_report_runtime() {
...
         foreach ($reporttypes as $reporttype) {
             $reportbuilderclass = "tool_objectfs\\local\\report\\{$reporttype}_report_builder";
             $reportbuilder = new $reportbuilderclass();
-            $report = $reportbuilder->build_report();
+            $report = $reportbuilder->build_report($reportstarted);
             objectfs_report_builder::save_report_to_database($report);
         }
+        // Throttle here for one second to make sure the snapshots have different
+        // $reportstarted if the report was called twice in a row and it was super fast.
+        sleep(1);
```
brendanheywood, line 122: This is a smell the design of the db table is not ideal

classes/local/report/object_location_history_table.php
golenkovm: Done. get_records is used

object_location.php
golenkovm: Added

object_location.php
golenkovm: Moved to the top

object_location.php
golenkovm: Pagesize is not being used here. I set it to 0. Hope this is not confusing now

classes/local/report/objectfs_report.php
golenkovm: Re-worked th edb schema, so there is no need to sleep now

classes/local/report/object_location_history_table.php
brendanheywood: This isn't done at all, it is still sing get_records, not get_recordset. This is a much deeper change

Issue #326, opened 2020-05-25
Allow no credentials in config, use aws roles for s3


brendanheywood:

golenkovm:
Implemented

PR #328, opened 2020-06-03
Allow no credentials in config, use aws roles for s3

golenkovm:
This is an implementation of #326


golenkovm:
Tested in Moodle 3.3 and Moodle 3.9:

image

When enabled, check notification is not displayed as well as key and secret fields:

image

Issue #306, opened 2020-05-06
sizethreshold not being respected


jgupta:
Even though _sizethreshold_ is set to 1024, we see small files of size 1kb being served from S3.

brendanheywood:
Is that an exact rounding issue? Like if you set it to 5kb do all files 1-5 kb still get streamed through s3 after they have had time for cron to move them back?

jgupta:
I am under impression that sizethreshold takes input in KB. So, a configuration of 1024 should mean 1024KB i.e. 1MB. S3, should not be having files with a size of 1KB. Or is my assumptions incorrect, is sizethreshold in bytes?

brendanheywood:
Can you send me a screenshot of the 'Object location' table you get in this report:

admin/tool/objectfs/object_status.php

Has it pushed everything over?

brendanheywood:
I think there is a small bug in the help text of that setting, I think it is actually bytes and the default is 10240 = 10kb

jgupta:
If this is in bytes and not in KB, then it makes sense. Is there a way to bring back files smaller than sizethreshold?

![Screen Shot 2020-05-07 at 7 37 34 PM](https://user-images.githubusercontent.com/797815/81304153-4d08b100-909a-11ea-918a-ca1159e8e0ab.png)

brendanheywood:
Yes in theory you just change that threshold and after a few cron cycles it will have pulled stuff back down

brendanheywood:
We will implement this shortly too which should help you see where the smallest files are:

https://github.com/catalyst/moodle-tool_objectfs/issues/309

jgupta:
Great. I changed sizethreshold to 10240 yesterday and I see now that there are many small files been duplicated in both filedir and external storage.

![Screen Shot 2020-05-08 at 9 24 04 AM](https://user-images.githubusercontent.com/797815/81368650-e91fd000-910d-11ea-876c-a5a65c9c8ffe.png)

Now bumping sizethreshold to 102400 for #308. #309 would be wonderful.

I noticed that I am still getting very small size pre-signed URLs (~1kb), I am hoping that since preferexternal is set to 0, Moodle should prefer files present in filedir. Should I open a new thread for this?

PR #329, opened 2020-06-03
Fix lang string for sizethreshold and presignedminfilesize settings

golenkovm:
Fix for #306


lang/en/tool_objectfs.php
```diff
@@ -148,7 +148,7 @@
 $string['settings:openstack:container_help'] = 'The name of the container that will store the objects.';
 
 $string['settings:filetransferheader'] = 'File Transfer Settings';
-$string['settings:sizethreshold'] = 'Minimum size threshold (KB)';
+$string['settings:sizethreshold'] = 'Minimum size threshold';
```
brendanheywood, line 151: I think this should have the units just the correct one

golenkovm: Done

Issue #321, opened 2020-05-12
Add other common doc types to the test page: eg odt


brendanheywood:
* [x] libreoffice writer
* [x] libreoffice spreadsheet
* [ ] ?

PR #331, opened 2020-06-04
Add ods and odt doc types to the test page.

golenkovm:
This is an implementation of #321


golenkovm:
I've added two files. When navigating to testing page I have 4 files downloaded automatically once the page is open. Signing mehtod: S3.

image

Issue #311, opened 2020-05-08
Settings should be in a navigation tree


dmitriim:
Currently all pages releated to the plugin are added to "tool" branch of the admin navigation. They are all needs to be under the plugin category like:

![image](https://user-images.githubusercontent.com/4626431/81369403-aa057500-9135-11ea-945c-9ae1cc8ca5db.png)

golenkovm:
Done. Closing this off

PR #333, opened 2020-06-05
Fix broken URL on Pre-Signed URL testing page and lang string

golenkovm:
An additional patch for #311 
- Fix the link that leads to Objectfs settings from Pre-Signed URL Testing page when Pre-Signed URLs are not supported by chosen storage file system.
- Fix lang string.


golenkovm:
image

Issue #335, opened 2020-06-09
Investigate stream to tmp file potential issue for files over 2meg


brendanheywood:
There is an issue with s3 streams from the AWS api not being fully seekable, so internally php transparently downloads the entire stream to a tmp file in order to seek over it.

`Read only streams are truly streaming by default and will not allow you to seek. This is because data read from the stream is not kept in memory or on the local filesystem. You can force a "r" stream to be seekable by setting the "seekable" stream context option true. This will allow true streaming of data from Amazon S3, but will maintain a buffer of previously read bytes in a 'php://temp' stream to allow seeking to previously read bytes from the stream.`

https://github.com/catalyst/moodle-local_aws/blob/master/sdk/Aws/S3/StreamWrapper.php#L19-L25

`php://temp will use a temporary file once the amount of data stored hits a predefined limit (the default is 2 MB). The location of this temporary file is determined in the same way as the sys_get_temp_dir() function.`

https://www.php.net/manual/en/wrappers.php.php

These are not visible from inside php and probably live somewhere like:

/tmp/systemd-private-d13ee8ffdaf84a039646d023fd03d636-apache2.service-NkD8xt/tmp/

This means every range request for a large file ends up copying the entire file multiple times over to get the full file served

A rough but working code snippet:

https://gist.github.com/durzo/029ba48d7bf25a920c3a227c70d7d131

golenkovm:
Implemented in #340

PR #340, opened 2020-06-17
Proxy range requests

golenkovm:
Issue #335 and #294 
* Implement get_filesize_by_contenthash() file system method.
* Introduce unit test case for manager class.
* Tweak get_header() method and add unit tests.
* Allow associative arrays to be passed to get_header() method.
* Add a patch for Moodle 3.3/Totara 12.
* Introduce Proxy range request feature.
* Add test range request on setting form.
* Add unit tests covering Proxy range request methods.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -412,6 +412,7 @@ abstract class object_file_system extends \file_system_filedir {
      * @param stored_file $file The file to send
      * @return bool success
      * @throws \dml_exception
+     * @throws \coding_exception
      */
     public function xsendfile_file(stored_file $file): bool {
         $contenthash = $file->get_contenthash();
@@ -421,6 +422,13 @@ abstract class object_file_system extends \file_system_filedir {
 
             return $this->redirect_to_presigned_url($contenthash, headers_list());
         }
+
+        if ($this->externalclient->support_presigned_urls() &&
+                $ranges = $this->get_valid_http_ranges($file->get_filesize())) {
+
+            return $this->externalclient->proxy_range_request($file, $ranges);
+        }
+
         return false;
     }
 
@@ -897,5 +905,32 @@ abstract class object_file_system extends \file_system_filedir {
     public function get_filesize_by_contenthash($contenthash) {
         global $DB;
         return $DB->get_field('files', 'filesize', ['contenthash' => $contenthash], IGNORE_MULTIPLE);
+    }
+
+    /**
+     * Gets valid HTTP ranges for range request.
+     * Throttles request length down to 5MB size if it's greater.
+     *
+     * @param  int          $filesize Size of the file to be served.
+     * @return object|false           Array of range
+     */
+    public function get_valid_http_ranges($filesize) {
+        $range = manager::get_header($_SERVER, 'HTTP_RANGE');
+        if (!empty($range)) {
+            preg_match('{bytes=(\d+)?-(\d+)?(,)?}i', $range, $matches);
+            if (empty($matches[3])) {
+                $ranges = new \stdClass();
+                $ranges->rangefrom = (isset($matches[1])) ? intval($matches[1]) : 0;
+                $ranges->rangeto = (isset($matches[2])) ? intval($matches[2]) : ($filesize - 1);
+                $ranges->length = $ranges->rangeto - $ranges->rangefrom + 1;
+                if ($ranges->length > 5242880) {
+                    // Stream files in 5MB-chunks.
+                    $ranges->rangeto = $ranges->rangefrom + 5242880 - 1;
+                    $ranges->length = 5242880;
+                }
+                return $ranges;
+            }
+        }
+        return false;
     }
 }
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -773,7 +773,7 @@ abstract class object_file_system extends \file_system_filedir {
      * @throws \dml_exception
      */
     public function presigned_url_should_redirect($contenthash, $headers = array()) {
-        global $DB, $_SERVER;
+        global $_SERVER;
 
         // Redirect regardless.
         if ($this->externalclient->presignedminfilesize == 0 &&
@@ -799,13 +799,8 @@ abstract class object_file_system extends \file_system_filedir {
         }
 
         // Redirect when the file size is bigger than presignedminfilesize setting.
-        $sql = 'SELECT MAX(filesize) as filesize
-                  FROM {files}
-                 WHERE contenthash = :contenthash
-                   AND filesize > :filesize';
-        $record = $DB->get_record_sql($sql, ['contenthash' => $contenthash, 'filesize' => 0]);
-
-        return ($record->filesize >= $this->externalclient->presignedminfilesize);
+        $filesize = $this->get_filesize_by_contenthash($contenthash);
+        return ($filesize >= $this->externalclient->presignedminfilesize);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


classes/local/store/object_file_system.php
```diff
@@ -898,4 +913,68 @@ public function get_filesize_by_contenthash($contenthash) {
...
+     */
+    public function serve_range_request_via_presigned_url($contenthash, $filesize, $ranges) {
+        $url = $this->externalclient->generate_presigned_url($contenthash, headers_list());
+        $headers = array(
+            'HTTP/1.1 206 Partial Content',
+            'Content-Length: '.$ranges->length,
+            'Range: bytes=' . $ranges->rangefrom . '-' . $ranges->rangeto,
+        );
+
+        $curl = curl_init();
```
brendanheywood, line 959: please avoid using the raw php curl libs, use the moodle lib:

$curl = new curl();

This is generally preferred, but if you are already using guzzle for some reasons then its ok as long as the moodle proxy options are honored

classes/local/store/object_file_system.php
```diff
@@ -898,4 +913,68 @@ public function get_filesize_by_contenthash($contenthash) {
...
+        curl_setopt($curl, CURLOPT_CONNECTTIMEOUT, 15);
+        curl_setopt($curl, CURLOPT_TIMEOUT, 15);
+        $content = curl_exec($curl);
+        curl_close($curl);
+
+        if ($content === false) {
+            return false;
+        } else {
+            header('HTTP/1.1 206 Partial Content');
+            header('Content-Range: bytes ' . $ranges->rangefrom . '-' . $ranges->rangeto . '/' . $filesize);
```
brendanheywood, line 973: Technically it is possible for the range that was requested to not be the range that was returned. Can you please grab these headers from the curl response and pass them through as-is

classes/local/store/object_file_system.php
```diff
@@ -421,6 +421,13 @@ public function xsendfile_file(stored_file $file): bool {
 
             return $this->redirect_to_presigned_url($contenthash, headers_list());
         }
+
+        if ($this->externalclient->support_presigned_urls()) {
+            $ranges = $this->get_valid_http_ranges($file->get_filesize());
+            if ($file->get_filesize() > 2097152 && $ranges) {
```
brendanheywood, line 427: can you make a const for this magic number and add a comment explaining it

brendanheywood: probably point to the php max temp / aws sdk comments

classes/local/store/object_file_system.php
```diff
@@ -443,6 +450,14 @@ public function xsendfile($contenthash) {
 
             return $this->redirect_to_presigned_url($contenthash, $headers);
         }
+
+        if ($this->externalclient->support_presigned_urls()) {
+            $filesize = $this->get_filesize_by_contenthash($contenthash);
+            $ranges = $this->get_valid_http_ranges($filesize);
+            if ($filesize > 2097152 && $ranges) {
+                $this->serve_range_request_via_presigned_url($contenthash, $file->get_filesize(), $ranges);
```
brendanheywood, line 458: is this a bug?

                $this->serve_range_request_via_presigned_url($contenthash, $file->get_filesize(), $ranges);

vs

                $this->serve_range_request_via_presigned_url($contenthash, $filesize, $ranges);

classes/local/store/object_file_system.php
golenkovm: `new curl();` has handy features indeed. Thanks.
Done.

classes/local/store/object_file_system.php
golenkovm: Done

classes/local/store/object_file_system.php
golenkovm: Made a constant and added a comment with a link.
Didn't find how to get this php max temp on the fly. 
Looks like it's 2MB by default, but can be overriden via code. And there is no way to get current `maxmemory`.

classes/local/store/object_file_system.php
golenkovm: Yep. Thanks. Fixed

classes/local/store/object_file_system.php
```diff
@@ -891,4 +901,105 @@ private function exec_command($command) {
...
+     * @param  stored_file $file    The file to send
+     * @param  object      $ranges  Object with rangefrom, rangeto and length properties.
+     * @return false if couldn't get data.
+     * @throws \coding_exception
+     */
+    public function serve_range_request(stored_file $file, $ranges) {
+        $response = $this->curl_range_request_to_presigned_url($file->get_contenthash(), $ranges, headers_list());
+        $httpcode = manager::get_header($response['responseheaders'], 'HTTP/1.1');
+        if ($response['content'] == '' || $httpcode != '206 Partial Content') {
+            debugging('Range request failed with HTTP code: ' . $httpcode . '. Details: ' . $response['content']);
```
brendanheywood, line 955: can you log the failed curl url here please

classes/local/store/object_file_system.php
```diff
@@ -421,6 +429,13 @@ public function xsendfile_file(stored_file $file): bool {
 
             return $this->redirect_to_presigned_url($contenthash, headers_list());
         }
+
+        if ($this->externalclient->support_presigned_urls()) {
```
brendanheywood, line 433: as discussed move this logic into serve_range_request() which will be in each store

golenkovm: Done

classes/local/store/object_file_system.php
golenkovm: Done

classes/local/store/s3/client.php
```diff
@@ -37,6 +37,14 @@
 define('AWS_CAN_DELETE_OBJECT', 2);
 
 class client extends object_client_base {
+
+    /**
+     * @var int A predefined limit of data stored.
+     * When hit, php://temp will use a temporary file.
+     * Reference: https://www.php.net/manual/en/wrappers.php.php
```
brendanheywood, line 44: Can you please link to this:

https://github.com/catalyst/moodle-local_aws/blob/master/sdk/Aws/S3/StreamWrapper.php#L19-L25

classes/local/store/s3/client.php
```diff
@@ -635,4 +643,71 @@ public function define_client_check_sdk($config) {
...
+        }
+
+        $response = $this->curl_range_request_to_presigned_url($file->get_contenthash(), $ranges, headers_list());
+        $httpcode = manager::get_header($response['responseheaders'], 'HTTP/1.1');
+
+        if ($response['content'] != '' && $httpcode == '206 Partial Content') {
+            header('HTTP/1.1 206 Partial Content');
+            header('Accept-Ranges: bytes');
+            $contentrange = manager::get_header($response['responseheaders'], 'Content-Range');
+            if ($contentrange !== '') {
```
brendanheywood, line 669: I think these 2 if statements could be simpler, it should always trust the headers from the s3 response. If those headers don't exist then don't fall back to $contentrange, just treat that the same as the other types of failures if the code is not 206. If they don't have these headers they are a totally invalid response

https://tools.ietf.org/html/rfc7233#section-4.1

> If a single part is being transferred, the server generating the 206 response MUST generate a Content-Range header field

classes/local/store/object_file_system.php
```diff
@@ -773,7 +773,7 @@ public function presigned_url_should_redirect_file($file) {
      * @throws \dml_exception
      */
     public function presigned_url_should_redirect($contenthash, $headers = array()) {
-        global $DB, $_SERVER;
+        global $_SERVER;
```
brendanheywood, line 784: not needed?

golenkovm: Yeah

classes/local/store/s3/client.php
```diff
@@ -635,4 +643,63 @@ public function define_client_check_sdk($config) {
...
+     * Proxy range request.
+     *
+     * @param  \stored_file $file    The file to send
+     * @param  object       $ranges  Object with rangefrom, rangeto and length properties.
+     * @return false If couldn't get data.
+     * @throws \coding_exception
+     */
+    public function proxy_range_request(\stored_file $file, $ranges) {
+        // Do not serve files with size less than 2MB or if the feature is disabled.
+        if ($file->get_filesize() < self::MAX_TEMP_LIMIT || empty($this->config->proxyrangerequests)) {
```
brendanheywood, line 658: can you swap the order of this so if the config is off it exist earlier

get_filesize can do some db / api calls in same cases

Issue #352, opened 2020-06-30
Add travis tests for 'is_ready' and nothing at all should happen if we are not ready


brendanheywood:
if if no store is configured, or if a store is configured but has not settings then it is not ready. If it is not ready then the whole code should behave exactly as the core filedir store

We need to test at least
* [ ] getting a file
* [ ] xsendfile

golenkovm:
Implemented as part of #354

PR #354, opened 2020-07-03
Introduce 'is_configured()' filesystem check, optimize setting page

golenkovm:
This is a fix for #352

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -357,16 +357,20 @@ abstract class object_file_system extends \file_system_filedir {
      * @return void
      */
     public function readfile(\stored_file $file) {
-        $path = $this->get_remote_path_from_storedfile($file);
+        if (!$this->is_configured()) {
+            parent::readfile($file);
+        } else {
+            $path = $this->get_remote_path_from_storedfile($file);
 
-        $this->logger->start_timing();
-        $success = readfile_allow_large($path, $file->get_filesize());
-        $this->logger->end_timing();
+            $this->logger->start_timing();
+            $success = readfile_allow_large($path, $file->get_filesize());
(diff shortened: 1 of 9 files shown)


classes/local/store/object_file_system.php
```diff
@@ -357,16 +357,20 @@ public function delete_empty_dirs($rootpath = '') {
      * @return void
      */
     public function readfile(\stored_file $file) {
-        $path = $this->get_remote_path_from_storedfile($file);
+        if ($this->is_configured()) {
```
brendanheywood, line 360: wondering if this would read better as

```
if ($this->is_configured()) {
    return parent::readfile($file);
}

...
```

we can repeat this pattern in a few places, ie supports_xsendfile should be the same (ie it can do a xsendfile for filedir if it is turned on)

golenkovm: I think you meant
```
if (!$this->is_configured()) {
    return parent::readfile($file);
} else {
   //objectfs stuff
}
```
`if (!$this->is_configured()) {` instead of `if ($this->is_configured()) {`

classes/local/store/object_file_system.php
```diff
@@ -415,21 +419,25 @@ public function get_content(\stored_file $file) {
      * @throws \coding_exception
      */
     public function xsendfile_file(stored_file $file): bool {
-        $contenthash = $file->get_contenthash();
-        if ($this->presigned_url_configured() &&
+        if (!$this->is_configured()) {
+            return parent::xsendfile_file($file);
+        } else {
```
brendanheywood, line 424: why not just ditch the else completely? we exit above. same for the other methods

golenkovm: Done

Issue #353, opened 2020-07-03
The expiration date of a signature version 4 presigned URL must be less than one week


golenkovm:
Not sure that there is a fix for this, but will record it. 
With some SCORM activities Moodle sets Expire to 3 months and therefore generating pre-signed url fails with error:

```
The expiration date of a signature version 4 presigned URL must be less than one week in #012* line 745 of /admin/tool/objectfs/classes/local/store/object_file_system.php: call to debugging()#012* line 424 of /admin/tool/objectfs/classes/local/store/object_file_system.php: call to tool_objectfs\local\store\object_file_system->redirect_to_presigned_url()#012* line 1836 of /lib/filestorage/file_storage.php: call to tool_objectfs\local\store\object_file_system->xsendfile_file()#012* line 2172 of /lib/filelib.php: call to file_storage->xsendfile_file()#012* line 2523 of /lib/filelib.php: call to readfile_accel()#012* line 2668 of /lib/filelib.php: call to send_file()#012* line 1019 of /mod/scorm/lib.php: call to send_stored_file()#012* line 4961 of /lib/filelib.php: call to scorm_pluginfile()#012* line 44 of /pluginfile.php: call to file_pluginfile()" while reading response header from upstream, client: 203.194.45.242, server: lms-uat.monash.edu, request: "GET /pluginfile.php/10807058/mod_scorm/content/1/mobile/5sn5qGfGezN_DX1258_DY1258_CX643_CY600.png HTTP/1.1", upstream: "fastcgi://unix:/var/run/php/php7.2-fpm.sock:", host: "xxx", referrer: "https://xxx/pluginfile.php/10807058/mod_scorm/content/1/index_lms.html"
```

brendanheywood:
see duplicate of https://github.com/catalyst/moodle-tool_objectfs/issues/360

PR #361, opened 2020-07-22
If expire header is greater than a week set it to 1 week.

golenkovm:
This is a fix for #353

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -570,16 +570,20 @@ class client extends object_client_base {
         if ($expires == 0) {
             $expires = $now + $this->expirationtime;
         }
-        // If it's already expired set it to now.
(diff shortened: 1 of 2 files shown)


classes/local/store/s3/client.php
```diff
@@ -570,16 +570,18 @@ public function get_expiration_time($now, $expires) {
         if ($expires == 0) {
             $expires = $now + $this->expirationtime;
         }
-        // If it's already expired set it to now.
+        // If it's already expired set it to 1 minute.
         if ($expires < $now) {
-            $expires = $now;
+            $expires = $now + MINSECS;
+        }
+        // If it's greater than a week set it to 1 week.
```
brendanheywood, line 577: I think it's important to say why we are doing this in the comment

golenkovm: Added

PR #363, opened 2020-07-24
Fix navigation buttons in Object status report.

golenkovm:
This is a fix for #364


object_status.php
```diff
@@ -33,13 +33,9 @@
 use tool_objectfs\local\report\object_status_history_table;
 
 $reportid = optional_param('reportid', 0, PARAM_INT);
```
brendanheywood, line 35: what happens when the param is missing?

golenkovm: Then `$reportid` is set ot 0. And later set to the latest report snapshot `$reportid = key($reports);`

Issue #365, opened 2020-07-27
unlink doesn't seem to delete files in S3 compatible storages


abrahammartin:
We are using GCP S3 compatible interface and files are not being deleted. 

On https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/s3/client.php#L173 unlink is used to delete the file for S3 storage:

```php
    /**
     * Deletes a file in S3 storage.
     *
     * @path   string full path to S3 file.
     */
    public function delete_file($fullpath) {
        unlink($fullpath);
    }
```

On https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/azure/client.php#L130 the azure client library call is used to delete the file for Azure storage:

```php
    /**
     * Deletes file (blob) in azure blob storage.
     *
     * @param string $fullpath path to azure blob.
     */
    public function delete_file($fullpath) {
        $container = $this->container;
        $relativepath = $this->get_relative_path_from_fullpath($fullpath);

        $this->client->deleteBlob($container, $relativepath);
    }
```

Shouldn't the function `delete_file` for S3 storage use the S3 library call to delete the file? This is used when testing deletion permissions in https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/s3/client.php#L276

```php
        if ($testdelete) {
            try {
                $result = $this->client->deleteObject(array('Bucket' => $this->bucket, 'Key' => 'permissions_check_file'));
                $permissions->messages[get_string('settings:deletesuccess', 'tool_objectfs')] = 'warning';
                $permissions->success = false;
            } catch (\Aws\S3\Exception\S3Exception $e) {
                $errorcode = $e->getAwsErrorCode();
                // Something else went wrong.
                if ($errorcode !== 'AccessDenied') {
                    $details = $this->get_exception_details($e);
                    $permissions->messages[get_string('settings:deleteerror', 'tool_objectfs') . $details] = 'notifyproblem';
                    $permissions->success = false;
                }
            }
        }
```

So I think `delete_file` for S3 storage should be:

```php
    /**
     * Deletes a file in S3 storage.
     *
     * @path   string full path to S3 file.
     */
    public function delete_file($fullpath) {
        $this->client->deleteObject(array('Bucket' => $this->bucket, 'Key' => $fullpath));
    }
```

syn:
We have intentionally not fully implemented deletes in S3 Buckets, 
[...]

PR #366, opened 2020-07-27
Use S3 client native functions to delete a S3 file

abrahammartin:
Use S3 client native functions to delete a S3 file instead of using unlink.

Proposed solution for https://github.com/catalyst/moodle-tool_objectfs/issues/365

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -170,7 +170,9 @@ class client extends object_client_base {
      * @path   string full path to S3 file.
      */
     public function delete_file($fullpath) {
-        unlink($fullpath);
+        $relativepath = $this->get_relative_path_from_fullpath($fullpath);
+
+        $this->client->deleteObject(array('Bucket' => $this->bucket, 'Key' => $relativepath));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


classes/local/store/s3/client.php
```diff
@@ -170,7 +170,9 @@ public function get_trash_fullpath_from_hash($contenthash) {
      * @path   string full path to S3 file.
      */
     public function delete_file($fullpath) {
-        unlink($fullpath);
+        $relativepath = $this->get_relative_path_from_fullpath($fullpath);
+
+        $this->client->deleteObject(array('Bucket' => $this->bucket, 'Key' => $relativepath));
```
brendanheywood, line 175: This looks ok at face value, but can you explain why this is better that the stream wrapper unlink which does basically exactly the same thing?

https://github.com/catalyst/moodle-local_aws/blob/master/sdk/Aws/S3/StreamWrapper.php#L216-L225

tuanngocnguyen:
Hi @abrahammartin 

Thanks for your PR.
Since this has been inactive for some time, I am closing it.
Please create new one against MOODLE_33_STABLE/MOODLE_310_STABLE if you are still working on the issue.

Regards

Issue #375, opened 2020-08-19
Correctly handle intermittent issues connecting to s3


brendanheywood:
There is possibly 2 issues that overlap and make things worse.

The first issue is there is some intermittent reason why it cannot connect to s3. This is the root cause and being addressed elsewhere, but when it can't read a number of knock on effects happen:

1) it can't read the file so the object ends up being set to OBJECT_LOCATION_ERROR

https://github.com/catalyst/moodle-tool_objectfs/blob/59a8783739b7e6a4865a56f46a030436cb71d96e/classes/local/store/object_file_system.php#L215-L225

This then heals later based on the cron recover task

2) The file still ends up being served as a 200 which in many cases then gets cached in the browser. This must throw an exception and end up as a 40x or 50x so it will not be cached.

PR #376, opened 2020-08-21
Implement additional check if the file is missing externally.

golenkovm:
#375

The logic is
Use `is_readable()` to check if the file is readable externally.
In most cases it will be readable.
If it's not we make a deeper check which will:
     * Return `true` if the file is readable.
     * Return `false` if the file is missing externally.
     * Throw an exceptions if there is any other issue with external storage.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -194,7 +194,15 @@ abstract class object_file_system extends \file_system_filedir {
         $path = $this->get_external_path_from_hash($contenthash, false);
 
         // Note - it is not possible to perform a content recovery safely from a hash alone.
-        return is_readable($path);
+        if (is_readable($path)) {
+            // File is readable externally, so everything is good here.
+            return true;
+        } else {
+            // Otherwise make a deeper check.
+            // This is needed to figure out if the file is simply missing externally or
+            // there is an issue with external storage.
+            return $this->externalclient->is_file_readable_by_hash($contenthash);
+        }
     }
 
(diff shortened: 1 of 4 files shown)


classes/local/store/object_file_system.php
```diff
@@ -194,7 +194,15 @@ public function is_file_readable_externally_by_hash($contenthash) {
...
         // Note - it is not possible to perform a content recovery safely from a hash alone.
-        return is_readable($path);
+        if (is_readable($path)) {
+            // File is readable externally, so everything is good here.
+            return true;
+        } else {
+            // Otherwise make a deeper check.
+            // This is needed to figure out if the file is simply missing externally or
+            // there is an issue with external storage.
+            return $this->externalclient->is_file_readable_by_hash($contenthash);
```
brendanheywood, line 204: in the base implementation it simply does the exact same is_readable a second time which seems wasteful.

Wouldn't it be better to move all of this logic purely into the s3 store?

golenkovm: Perhaps simply return false in the base implementation?

tuanngocnguyen:
hi @golenkovm 

Thanks for your PR.


Since there are changes in the supported branches as per: https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/README.md#Branches, Would you please create new MR against MOODLE_33_STABLE/MOODLE_310_STABLE if you are still working on the issue?
Thanks

Regards

PR #395, opened 2021-01-17
Added checks to delete orphaned files in filedir

Peterburnett:
closes https://github.com/catalyst/moodle-tool_objectfs/issues/396

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -331,6 +331,10 @@ abstract class object_file_system extends \file_system_filedir {
      * @return bool
      */
     public function delete_empty_dirs($rootpath = '') {
+        global $DB;
+
+        $config = manager::get_objectfs_config();
+
         if (empty($rootpath)) {
             $rootpath = $this->filedir;
         }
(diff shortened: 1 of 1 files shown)


classes/local/store/object_file_system.php
```diff
@@ -339,19 +341,39 @@ public function delete_empty_dirs($rootpath = '') {
         }
         $empty = true;
         foreach (glob($rootpath . DIRECTORY_SEPARATOR . '*') as $path) {
-            // If there are .tmp files here, they should be killed.
-            if (!is_dir($path) && pathinfo($path, PATHINFO_EXTENSION) === 'tmp') {
+            if (!is_dir($path)) {
```
brendanheywood, line 344: wonder if this should be is_file instead?

classes/local/store/object_file_system.php
```diff
@@ -339,19 +341,39 @@ public function delete_empty_dirs($rootpath = '') {
...
+                    // If there are .tmp files here, they should be killed.
+                    if (!empty($pathinfo['extension']) && $pathinfo['extension'] === 'tmp') {
+                        @unlink($path);
+                    } else {
+                        // If the file is 24hrs old, they may not be tracked by objectFS.
+                        // This means it  may not exist in the files table at all.
+
+                        // Check the basename and filename, should catch hidden files and other junk.
+                        $exists = (
+                            $DB->record_exists('files', ['contenthash' => $pathinfo['filename']]) ||
```
brendanheywood, line 363: Is it worth grouping these into a single sql query?

classes/local/store/object_file_system.php
```diff
@@ -339,19 +341,39 @@ public function delete_empty_dirs($rootpath = '') {
...
-                    @unlink($path);
+                if ($modified <= (time() - DAYSECS)) {
+                    $pathinfo = pathinfo($path);
+
+                    // If there are .tmp files here, they should be killed.
+                    if (!empty($pathinfo['extension']) && $pathinfo['extension'] === 'tmp') {
+                        @unlink($path);
+                    } else {
+                        // If the file is 24hrs old, they may not be tracked by objectFS.
+                        // This means it  may not exist in the files table at all.
```
brendanheywood, line 359: double space in comment

Also wondering about performance if you have chosen to set the sitethreshold setting, we will constantly keep doing db calls grinding through them. Not sure best approach here

Peterburnett: To combat this, I have avoided any files that are underneath the size threshold. It will still do a little bit of comparison on them, but nowhere near as heavy as a db call to check them out (plus it only runs once a week). You get very little benefit from the tiny files anyway, unless there are millions of them.

This issue also appears if deletelocal is disabled, as the filedir will contain massive amounts of files. In this instance, I've opted just to skip the checks, as the likely improvement is only a few %, and it will take a massive amount of time to process.

classes/local/store/object_file_system.php
```diff
@@ -339,19 +343,58 @@ public function delete_empty_dirs($rootpath = '') {
...
-                if ($delete) {
-                    @unlink($path);
+                if ($modified <= (time() - DAYSECS)) {
+                    $pathinfo = pathinfo($path);
+
+                    // If there are .tmp files here, they should be killed.
+                    if (!empty($pathinfo['extension']) && $pathinfo['extension'] === 'tmp') {
+                        @unlink($path);
+                    } else {
+                        if (!$config->deletelocal) {
```
brendanheywood, line 360: shouldn't this be way up at the top of the loop?

brendanheywood:
This looks good to merge

PR #397, opened 2021-01-31
Fix issues with swift exceptions.

### LINKED ISSUE #382: delete_local_objects task failing on 3.9 site using SWIFT
Scheduled task failed: Object file system delete local objects task (tool_objectfs\task\delete_local_objects),Coding error detected, it must be fixed by a programmer: A lock was created but not released at:

/admin/tool/objectfs/classes/local/store/object_file_system.php on line 235

 Code should look like:

 $factory = \core\lock\lock_config::get_lock_factory('type');
 $lock = $factory->get_lock(113294202);
 $lock->release();  // Locks must ALWAYS be released like this.

Backtrace:
* line 108 of /admin/tool/objectfs/classes/local/object_manipulator/manipulator.php: call to core\lock\lock->__destruct()
* line 81 of /admin/tool/objectfs/classes/local/object_manipulator/manipulator_builder.php: call to tool_objectfs\local\object_manipulator\manipulator->execute()
* line 66 of /admin/tool/objectfs/classes/task/task.php: call to tool_objectfs\local\object_manipulator\manipulator_builder->execute()
* line 248 of /lib/cronlib.php: call to tool_objectfs\task\task->execute()
* line 120 of /lib/cronlib.php: call to cron_run_inner_scheduled_task()
* line 73 of /lib/cronlib.php: call to cron_run_scheduled_tasks()
* line 79 of /admin/cli/cron.php: call to cron_run()

danmarsden:
Fixes #382

This fixes a bunch of issues for us on Catalyst Cloud (openstack)

### DIFF
--- a/classes/local/store/swift/client.php
+++ b/classes/local/store/swift/client.php
@@ -137,7 +137,7 @@ class client extends object_client_base {
 
             $obj->retrieve();
 
-        } catch (BadResponseError $e) {
+        } catch (\OpenStack\Common\Error\BadResponseError $e) {
             return false;
         }
 
@@ -179,7 +179,7 @@ class client extends object_client_base {
 
         try {
             $container->createObject(['name' => 'connection_check_file', 'content' => 'connection_check_file']);
-        } catch (BadResponseError $e) {
+        } catch (\OpenStack\Common\Error\BadResponseError $e) {
             $connection->success = false;
             $connection->details = $this->get_exception_details($e);
         } catch (Exception $e) {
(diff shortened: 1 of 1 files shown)


danmarsden:
note - I considered adding an exception handler rather than replace the existing /badresponse one, but It seemed more correct to replace it with the full openstack path to the error.

Would be interested to hear if someone thinks we need to leave the /badresponse in there as well as the full path to the openstack badresponse.

Peterburnett:
The change itself seems fine, but it seems this has the potential to create a divergence in the supported openstack libs.

I wonder if it should be something a little more generic like

```
catch (Exception $e) {
    if ($e instanceof <NEW> || $e instaceof <OLD>) {
        do...
    } else {
        throw $e
    }
}
```

danmarsden:
yeah - I was assuming that the old "BadResponseError" was in fact incorrect but might have been "working" in php 5 - I guess I should go digging in old openstack libs - or we should make the current openstack lib a dependency on this version.
I'll check to make sure before merging this in.

danmarsden:
looking through old code this just seems to be a copy/paste error - copying from openstack code but missing: 
use OpenStack\Common\Error\BadResponseError

from the tops of the files - I also wonder if the changes to exceptions in PHP 7 meant that when running this on php 5 we didn't notice the issue in the same way.

PR #399, opened 2021-02-05
Only peform directory clean up on scheduled task.

### LINKED ISSUE #398: Empty dir cleanup should only run on scheduled task
delete_local_file_from_hash and delete_object_from_local_by_hash function will run an empty directory cleanup after a file has been deleted.

This directory clean up should really only happen on scheduled task - this is even more important for Catalyst NZ as we use one layer storage (most files are stored in filedir/root rather than filedir/XX/AA/FILE.) and the way the dir_cleanup function is written takes a long time to iterate and find stuff to delete.

We want to avoid unnessary IO tasks on the users session - cleanup tasks are best run on scheduled  task.

danmarsden:
Fix #398  
@Peterburnett do you want to review this one?

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -310,8 +310,6 @@ abstract class object_file_system extends \file_system_filedir {
                 $success = unlink($localpath);
 
                 if ($success) {
-                    // Check grandparent dir for empty dirs.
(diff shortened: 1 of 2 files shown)


brendanheywood:
I'm late to the party but this patch feels sub optimal. Detecting whether a dir is empty of not, should cost a single stat call, we shouldn't have removed this because it was expensive, we should have optimized it.

PR #403, opened 2021-02-05
If deletelocal is false, don't run empty dir cleanup either.

### LINKED ISSUE #400: Empty dir clean up should not run if "deletelocal" is set to false.
At the moment we have the setting "deletelocal" which suggests that nothing local will be deleted...

but then empty dir clean up runs and quietly deletes some stuff - IMO when deletelocal is false, it shouldn't run that task at all.

danmarsden:
Fix #400

I haven't tested this one locally yet, so might be best to hold off until someone (me probably) puts this on a local site and makes sure I've got the logic right :-)

### DIFF
--- a/classes/task/delete_local_empty_directories.php
+++ b/classes/task/delete_local_empty_directories.php
@@ -26,6 +26,7 @@
 namespace tool_objectfs\task;
 
 use coding_exception;
+use tool_objectfs\local\manager;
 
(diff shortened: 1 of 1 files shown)


Peterburnett:
I think this one is good to go, The change is not destructive, and the logic is sound and simple. Happy to merge, but I will leave you to click the button @danmarsden if you did want to test first

PR #407, opened 2021-02-12
ISSUE-406: Fix error 'property of non-object'.

### LINKED ISSUE #406: Error filling up logs
There is an error caused by a recent patch, which is filling up the error logs on some of our clients very quickly:

`Trying to get property 'sizethreshold' of non-object.`

I will submit a pull request to fix this issue momentarily.

cwarwicker:
This was trying to get the property from $this->config instead of $config.

This fixes #406


brendanheywood:
thanks @cwarwicker for the quick patch

PR #323, opened 2020-05-12
fixed digitalocean settings issue

hohenstaufen:
digitalocean client was missing some settings that were present in the standard s3 implementation. This fixes #318

Issue #318, opened 2020-05-11
No candidate objects found


hohenstaufen:
Upload task always report "No candidate objects found".

`Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 17:23:06. Current memory use 36.6MB.
No candidate objects found.
... used 1 dbqueries
... used 0.18827605247498 seconds
Scheduled task complete: Object file system upload task (tool_objectfs\task\push_objects_to_storage)`

On moodle 3.8.3.
![settings](https://user-images.githubusercontent.com/1156803/81579812-e5f63f80-93ac-11ea-9adb-f2bcdbbbb1c1.png)

![immagine](https://user-images.githubusercontent.com/1156803/81579952-18a03800-93ad-11ea-80b4-c4bac2600581.png)

Any idea on what could be wrong?

dennis-ciancoders:
Any hints on this?, I'm having the exact same problem

choma:
I was having this problem too. I'm using moodle 3.8.2 with DigitalOcean Spaces.
It was correctly uploading all the 9 [tests files](https://github.com/catalyst/moodle-tool_objectfs/tree/master/tests/fixtures), but no other files were uploaded.

I applied the changes on PR #323 by hand, and it started to work. Is it safe to leave it like this? is the travis problem the only reason to not merge this PR?

guillefd:
Yes, this PR #323 is not merged to the stable version (don't know why) but it works ok, I'm using DO spaces on 4 production sites for 6 month now (on moodle 3.9), and works great.
The only detail is that files on spaces are not being deleted when they are deleted locally, have tried the last moodle-tool_objectfs version and still has this issue, but everything else works fine.

gcavalcante8808:
I just created the PR https://github.com/catalyst/moodle-tool_objectfs/pull/413, which is simpler than #323, as it apply the specific 323 DO client fix to the current master.

PR #413, opened 2021-03-28
fix: Add needed client options to work with digital ocean spaces

gcavalcante8808:
## Scenario

As stated in https://github.com/catalyst/moodle-tool_objectfs/issues/318, when we use the digital ocean spaces client, the diff for changed files isn't working properly.

Looking at the PR https://github.com/catalyst/moodle-tool_objectfs/pull/323, the real fix for this is really simple, so I made this PR to the minimal required change.

## What Has Been Done

- [x] Corrected DigitalOcean client setup.

### DIFF
--- a/classes/local/store/digitalocean/client.php
+++ b/classes/local/store/digitalocean/client.php
@@ -35,9 +35,16 @@ class client extends s3_client {
         $this->autoloader = $CFG->dirroot . '/local/aws/sdk/aws-autoloader.php';
         $this->testdelete = false;
 
+
         if ($this->get_availability() && !empty($config)) {
             require_once($this->autoloader);
             $this->bucket = $config->do_space;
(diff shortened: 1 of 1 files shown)


degodfather:
@gcavalcante8808 This worked like charm. Thank you!

classes/local/store/digitalocean/client.php
```diff
@@ -35,9 +35,16 @@ public function __construct($config) {
         $this->autoloader = $CFG->dirroot . '/local/aws/sdk/aws-autoloader.php';
         $this->testdelete = false;
 
+
```
brendanheywood, line 38: Can you please remove this blank line (breaks code standards ci) and this this should be good to land

gcavalcante8808: Done.

PR #415, opened 2021-04-02
ADD composer.json

michaelmeneses:


danmarsden:
closing as won't fix for now - we don't use composer to manage plugin installs, and this doesn't seem to have been widely adopted by the Moodle community either - happy to revisit in future, but this really needs to be documented in moodle's docs somewhere before then too.

PR #425, opened 2021-07-21
Remove clearing files dir when resetting between tests

dmitriim:


Peterburnett:
I just wonder if this will impact on any other tests like filelib_test.php, if there are files remaining in filedir after the tests. Perhaps the cleaner solution would be to set filedir to a specific folder like filedir/objectfs_test in the setUp() method and then delete just that dir in the teardown method

dmitriim:
Original issue with the failing tests doesn't have many details on how to replicate https://github.com/catalyst/moodle-tool_objectfs/issues/226 
I will try to run all file system related tests locally in 3.11 and see how it goes

dmitriim:
```
./vendor/bin/phpunit --stop-on-failure --testsuite core_files_testsuite,tool_objectfs_testsuiteite,core_files_testsuite
Moodle 3.11 (Build: 20210517), 8a960cc0ce2144424806db2f5c3b8a37ee7ebcc6
Php: 7.4.3, pgsql: 10.15 (Debian 10.15-1.pgdg90+1), OS: Linux 5.8.0-59-generic x86_64
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

...............................................................  63 / 261 ( 24%)
............................................................... 126 / 261 ( 48%)
............................................................... 189 / 261 ( 72%)
............................................................... 252 / 261 ( 96%)
.........                                                       261 / 261 (100%)

Time: 00:19.178, Memory: 97.50 MB

OK (261 tests, 1469 assertions)
root@9f740fe0efd9:/var/www/site# 

```

dmitriim:
will run with core_testsuite added

dmitriim:
@Peterburnett tests with core_testsuite are passing. It's good to go!

PR #429, opened 2021-09-01
Fix #428 - add ui setting for delete external.

### LINKED ISSUE #428: Convert $CFG->tool_objectfs_delete_externally to $CFG->forced_plugin_settings['tool_objectfs']['deleteexternal']
This should never have been implemented as:

$CFG->tool_objectfs_delete_externally;

it should have been a normal config item:

$CFG->forced_plugin_settings['tool_objectfs']['deleteexternal'] = true;

This should also have UI to show its state, even if its forced / locked

danmarsden:
#428


brendanheywood:
For anyone review this please don't merge until this internal wr is sorted:

https://wrms.catalyst.net.nz/wr.php?request_id=365407

PR #430, opened 2021-09-02
Add missing rename function for swift stream_wrapper.

danmarsden:
adds the missing rename function - 

 rename_external_file is called by object storage when a delete is triggered
that calls rename() function in swift client:
https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/swift/client.php#L320
which calls php function "rename()"
and "rename()" isn't in the stream wrapper:
https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/swift/stream_wrapper.php

### DIFF
--- a/classes/local/store/swift/stream_wrapper.php
+++ b/classes/local/store/swift/stream_wrapper.php
@@ -432,6 +432,33 @@ class stream_wrapper {
     }
 
     /**
+     * Rename object
+     *
+     * @param string $path
+     * @return boolean
+     */
+
+    public function rename($currentpath, $destinationpath) {
+        $currenturl = $this->parse_url($currentpath);
+        $destinationurl = $this->parse_url($destinationpath);
+
+        $options = ['destination' => $destinationurl['host'] . '/' . $destinationurl['path']];
+        try {
+            $this->get_container($currenturl['host'])->getObject($currenturl['path'])->copy($options);
+        } catch (\OpenStack\Common\Error\BadResponseError $e) {
+            trigger_error("Unable to copy file, errorcode:". $e->getResponse()->getStatusCode(), E_USER_WARNING);
+            return false;
+
+        } catch (\Exception $e) {
+            trigger_error("Unable to copy file", E_USER_WARNING);
+            return false;
+
+        }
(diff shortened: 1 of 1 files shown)


danmarsden:
still testing this - i'll drop into a testing site and update here once I can verify it's working.

Note - this also seems to fix #289

alexmorrisnz (approved):
I've tested this and can confirm it is working

classes/local/store/swift/stream_wrapper.php
```diff
@@ -431,6 +431,33 @@ public function unlink($path) {
...
+     * @return boolean
+     */
+
+    public function rename($currentpath, $destinationpath) {
+        $currenturl = $this->parse_url($currentpath);
+        $destinationurl = $this->parse_url($destinationpath);
+
+        $options = ['name' => $destinationurl['path'], 'containerName' => $destinationurl['host']];
+        try {
+            $this->get_container($currenturl['host'])->getObject($currenturl['path'])->copy($options);
```
alexmorrisnz, line 447: I think this needs to be changed to `['destination' => $pathhere]` as per https://php-openstack-sdk.readthedocs.io/en/latest/services/object-store/v1/objects.html#copy-object
Otherwise I get a 412 precondition failed destination header required
With that in place I get 403 forbidden, this looks to be something specific to the openstack system I'm testing with

alexmorrisnz: Line 445 needs to be `$options = ['destination' => $destinationurl['host'] . '/' . $destinationurl['path']];`
That will correctly move the file.

danmarsden:
Thanks @alexmorrisnz, I've pushed through an update for that - can you please confirm that looks correct and I'll merge it in.

alexmorrisnz (approved):

alexmorrisnz:
> Thanks @alexmorrisnz, I've pushed through an update for that - can you please confirm that looks correct and I'll merge it in.

Looks good :+1:

danmarsden:
the failures are unrelated and due to a new coding guildeline check (see #460) - merging this in to get it into our systems...

PR #433, opened 2021-09-22
T12 Unit text fix, test_delete_empty_folders()

nhoobin:
During PHPUnit execution the sitedata folder is not cleaned up with this test on T12.
> tool_objectfs\tests\object_file_system_testcase::test_delete_empty_folders


nhoobin:
The unit test failures for T12 were introduced with https://github.com/catalyst/moodle-tool_objectfs/pull/425

Issue #440, opened 2021-10-27
Orphaned records in tool_objectfs_objects create additional CPU load


ngmares:
On some larger sites the deleter and pusher candidate queries are responsible for quite a bit of load on the database. I've identified a large number of orphaned records in the mdl_tool_objectfs_objects table, i.e. records where there is no longer a matching contenthash in the mdl_files table. These orphaned records are substantially increasing the size of joins in these candidate queries which are run very regularly by cron. Because the records have been orphaned their state doesn't progress, and they are included in the join every time these queries are run by cron.

A mdl_tool_objectfs_object record will be orphaned where a file has been uploaded to Moodle, the file has been copied to external storage, e.g. S3, and then a user or some other process has deleted the file from within Moodle, resulting in the mdl_files record with the matching contenthash being deleted. I've run the following query on one large site which shows close to 2 million of these orphaned records in different states, with the bulk in 1 (duplicated) and 2 (remote only):

```
SELECT o.location, count(*)
FROM mdl_tool_objectfs_objects o
	left join mdl_files f ON o.contenthash = f.contenthash
WHERE f.id is null
	AND o.contenthash != 'da39a3ee5e6b4b0d3255bfef95601890afd80709' --empty files (e.g. ".")
GROUP BY o.location
```

If we were to archive these files by setting location to an "archived" state such as -2, then this would reduce the size of the joins in the deleter and pusher candidate queries.

On my site the deleter candidate query runs for 37 seconds before archiving any orphaned records. After archiving this same query runs in 1 second.

We would need to account for situations where an "archived" file is re-uploaded to moodle at a later date so the location can be updated to 1 (duplicated) or 0 (never copied to external storage) but this already looks to be covered adequately through the various calls to manager::get_object_location_from_hash() in the various add_file*** functions in class object_file_system.

PR #441, opened 2021-10-27
Add archiver task to better handle orphaned objects and improve perfo…

ngmares:
This pull request addresses implements an archiver task to address the performance/load issue detailed in this issue: https://github.com/catalyst/moodle-tool_objectfs/issues/440


brendanheywood:
This will be replaced by this pr https://github.com/catalyst/moodle-tool_objectfs/pull/451

Issue #447, opened 2021-11-22
Missing extra validation when getting links for missing files


guillogo:
Hi, when the links in the `Missing from filedir and external storage files` table for `component like mod_%` are being created, is assuming that all context level are modules. However, there are some cases where the context level is different, e.g.: `component = 'mod_hvp' and filearea = 'exports'` as these exports belong to course context level.

If there is a missing object with this particular case, the table will fail when tries to get the course from get_course_and_cm_from_cmid with a wrong course module id:

```
Error code: invalidrecordunknown
×Stack trace:
line 1682 of /lib/dml/moodle_database.php: dml_missing_record_exception thrown
line 2254 of /lib/modinfolib.php: call to moodle_database->get_record_sql()
line 116 of /admin/tool/objectfs/classes/local/table/files_table.php: call to get_course_and_cm_from_cmid()
line 797 of /lib/tablelib.php: call to tool_objectfs\local\table\files_table->col_link()
line 1920 of /lib/tablelib.php: call to flexible_table->format_row()
line 2044 of /lib/tablelib.php: call to table_sql->build_table()
line 51 of /admin/tool/objectfs/missing_files.php: call to table_sql->out()
```

PR #448, opened 2021-11-22
Issue #447 - Add additional validation when getting links for missing files

guillogo:
To fix: https://github.com/catalyst/moodle-tool_objectfs/issues/447

### DIFF
--- a/classes/local/table/files_table.php
+++ b/classes/local/table/files_table.php
@@ -40,7 +40,7 @@ class files_table extends \table_sql {
     public function __construct($uniqueid, $objectlocation) {
         parent::__construct($uniqueid);
 
-        $fields = 'f.*, ctx.instanceid';
+        $fields = 'f.*, ctx.instanceid, ctx.contextlevel';
         $from = '{files} f';
         $from .= ' LEFT JOIN {tool_objectfs_objects} o on f.contenthash = o.contenthash';
         $from .= ' LEFT JOIN {context} ctx ON f.contextid = ctx.id';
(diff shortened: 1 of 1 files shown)


classes/local/table/files_table.php
```diff
@@ -113,11 +113,19 @@ public function col_link(\stdClass $row) {
...
-            if (!empty($cm)) {
-                $url = new \moodle_url($cm->url);
+            switch ((int)$row->contextlevel) {
+                case CONTEXT_MODULE:
+                    list ($course, $cm) = get_course_and_cm_from_cmid($row->instanceid);
+                    if (!empty($cm)) {
+                        $url = new \moodle_url($cm->url);
+                    }
+                    break;
+                case CONTEXT_COURSE:
```
dmitriim, line 123: how could it be that $row->component contains mod, but $row->contextlevel would be CONTEXT_COURSE?

guillogo: Hi, there are some module type components where the context level is course or even system, for instance:

```
Component: mod_hvp
filearea: exports
context: 50
```

This record (mdl_files) is created each time a user creates a hvp activity with the option `Allow download` enabled; or when a user downloads a hvp activity if the previous record does not exits. The context in this case is 50 (course).

There are other module type components in mdl_files where the contextlevel is not CONTEXT_MODULE:


 

```
select distinct f.component, f.filearea, ctx.contextlevel
from mdl_context ctx
join mdl_files f ON f.contextid = ctx.id
where f.component like 'mod_%' and contextlevel <> 70;

e.g:

component,filearea,contextlevel
mod_turnitintool,submission,30
mod_data,site_presets,10
mod_hvp,exports,50
mod_hvp,libraries,10
mod_hvp,editor,50
mod_hvp,cachedassets,10
mod_hvp,library_archives,10
```

PR #450, opened 2021-11-23
Use newer CloudFront setup options

mpalimaka:


CLOUDFRONT.md
```diff
@@ -2,13 +2,11 @@
 1. Login to AWS console https://aws.amazon.com/console/
 2. Navigate to _Services -> S3_.
 3. Click _Create bucket_.
-4. Fill out bucket name, region and click _Create bucket_.
-5. Navigate to _My Security Credentials_.
-6. In the _Access keys_ section click on the _Create New Access Key_ button.
```
golenkovm, line 7: I believe we still need to create an access key.

CLOUDFRONT.md
```diff
@@ -2,13 +2,11 @@
...
-7. Write down your bucket name, region, key and secret.
-8. Edit the bucket again.
-9. Set _Default encryption_ to _Enabled_ with _Amazon S3 master-key (SSE-S3)_ server-side encryption.
-10. Set the following as _Cross-origin resource sharing (CORS)_:
+4. Fill out the bucket name and region
+5. Ensure _Block all public access_ is ticked
+6. Enable _Server-side encryption_ with _Amazon S3 key (SSE-S3)_
+7. Click _Create bucket_.
+8. Set the following as _Cross-origin resource sharing (CORS)_:
```
golenkovm, line 9: Should we set Cross-origin resource sharing and _then_ click Save changes?

CLOUDFRONT.md
```diff
@@ -58,51 +56,60 @@ php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\generate_status_
...
+6. Enter key value. Use the following command to get the public key:
 ```
 cat public_key.pem
 ```
-10. Click _Create public key_.
-11. Write down key ID from the [https://console.aws.amazon.com/cloudfront/v3/home#/publickey] page.
+7. Click _Create public key_.
+8. Write down key ID from the https://console.aws.amazon.com/cloudfront/v3/home#/publickey page.
+9. Navigate to https://console.aws.amazon.com/cloudfront/v3/home#/keygrouplist
+10. Create a key group and select the public key created previously
```
golenkovm, line 78: Yeah, this bit me and I still can feel the pain... Weird it didn't work with `self`

CLOUDFRONT.md
```diff
@@ -58,51 +56,60 @@ php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\generate_status_
...
-11. _Legacy cache settings -> Query strings_: Choose _All_.
-12. Click _Create distribution_.
-13. Navigate to [https://console.aws.amazon.com/cloudfront/v3/home?region=ap-southeast-2#/distributions].
-14. Confirm, that _Status_ is _Enabled_ and _Last modified_ is changed from _Deploying_ to the date the distribution was created.
-15. Open your distribution.
-16. Write down _Distribution domain name_ (with https://).
-> Note: If you have already setup Moodle behind a CloudFront distribution, it is also possible to use that same CloudFront distribution to serve files from objectfs. In this scenario, a specific prefix in the URL path directs traffic to the S3 Bucket (moodle.domain/objectfs/ for example). To achieve that, use the key_prefix option to add a prefix on your Bucket, and configure a second Origin on your existing CloudFront distribution that points to your Bucket. Setup a Behavior that uses that new Origin with the same prefix as the one you used as key_prefix in your Bucket. Follow all other instructions.
-
-### Generate keys
-1. Make a directory _$CFG->dataroot . '/objectfs/'_.
```
golenkovm, line 81: I'd leave this as objectfs can work with both file on the filesystem and a key stored as an admin setting

CLOUDFRONT.md
```diff
@@ -58,51 +56,60 @@ php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\generate_status_
...
+### Create CloudFront distribution
+1. Navigate to https://console.aws.amazon.com/cloudfront/.
+2. Click on _Create a CloudFront distribution_.
+3. Choose your Amazon S3 bucket from _Origin domain_ dropdown menu.
+4. _S3 bucket access_: Choose _Yes use OAI (bucket can restrict access to only CloudFront)_ and click _Create new OAI_.
+5. _S3 bucket access -> Bucket policy_: Choose _Yes, update the bucket policy_.
+6. _Viewer protocol policy_: Choose _Redirect HTTP to HTTPS_.
+7. _Allowed HTTP methods_: Choose _GET, HEAD, OPTIONS_ and tick _OPTIONS_ under _Cache HTTP methods_.
+8. _Restrict viewer access_: Choose _Yes -> Trusted key groups (recommended)_.
+9. _Cache key and origin requests_: Choose _Cache policy and origin request policy (recommended)_.
```
golenkovm, line 103: We'd need to add the key groups and choose the key group created above.

CLOUDFRONT.md
golenkovm: I mean this doc is for setting up all the evn from scratch, so we need an access key. If someone already have a bucket or use IAM Role they can simply skip this step

CLOUDFRONT.md
mpalimaka: The S3 interface has changed - CORS cannot be configured until after the bucket has been created.

CLOUDFRONT.md
mpalimaka: It is still possible to use `self`, but that is a legacy method (and requires root account access) and using key groups is now recommended instead.

CLOUDFRONT.md
mpalimaka: I have restored this and added text noting that two options are possible.

CLOUDFRONT.md
mpalimaka: Added addition text with this step.

CLOUDFRONT.md
mpalimaka: This text is confusing - S3 has no `Access keys` section and there's no reference to the other IAM steps required provide access to a bucket (but this is already detailed in `README.md`.

golenkovm: There is a step before to navigate to _My Security Credentials_. I had a look at `README.md` and I couldn't find details there on how to create Access keys. Could you please point me on them?

CLOUDFRONT.md
golenkovm: Ok, makes sense

CLOUDFRONT.md
```diff
@@ -58,29 +56,12 @@ php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\generate_status_
...
-16. Write down _Distribution domain name_ (with https://).
-> Note: If you have already setup Moodle behind a CloudFront distribution, it is also possible to use that same CloudFront distribution to serve files from objectfs. In this scenario, a specific prefix in the URL path directs traffic to the S3 Bucket (moodle.domain/objectfs/ for example). To achieve that, use the key_prefix option to add a prefix on your Bucket, and configure a second Origin on your existing CloudFront distribution that points to your Bucket. Setup a Behavior that uses that new Origin with the same prefix as the one you used as key_prefix in your Bucket. Follow all other instructions.
+### Generate CloudFront keys
+ObjectFS can use a CloudFront key from either the local filesystem or an admin setting. If using an admin setting, the key may be generated outside of the Moodle environment.
 
-### Generate keys
 1. Make a directory _$CFG->dataroot . '/objectfs/'_.
 2. Make it readable and writable:
-```
+ ```
```
golenkovm, line 64: Seems like an extra space in the beginning of the line

mpalimaka: Thanks, fixed.

golenkovm:
Thank you for the patch @mpalimaka

PR #451, opened 2021-12-01
Add archiver task to handle orphaned objects

### LINKED ISSUE #440: Orphaned records in tool_objectfs_objects create additional CPU load

keevan:
Closes #440 

Added reporting support, unit tests, typehints, comments and cleaned up the base implementation.

Screenshots of admin setting:

image

Few screenshots of the reporting:

image

image


classes/task/delete_archived_object_metadata.php
```diff
@@ -0,0 +1,56 @@
...
+    protected $stringname = 'delete_archived_object_metadata_task';
+
+    /**
+     * Execute task
+     */
+    public function execute() {
+        global $DB;
+
+        $wheresql = 'location = :location and timeduplicated < :ageforremoval';
+        $ageforremoval = 30 * DAYSECS;
```
brendanheywood, line 45: turn this into a setting which defaults to 0 which means don't ever delete

db/tasks.php
```diff
@@ -53,6 +53,24 @@
         'dayofweek' => '*',
         'month'     => '*'
     ),
+    array(
+        'classname' => 'tool_objectfs\task\archive_orphaned_objects',
+        'blocking'  => 0,
+        'minute'    => '52',
+        'hour'      => '*/12',
```
brendanheywood, line 60: make this once a day

db/tasks.php
```diff
@@ -53,6 +53,24 @@
...
+        'day'       => '*',
+        'dayofweek' => '*',
+        'month'     => '*'
+    ),
+    array(
+        'classname' => 'tool_objectfs\task\delete_archived_object_metadata',
+        'blocking'  => 0,
+        'minute'    => 'R',
+        'hour'      => '*/12',
```
brendanheywood, line 69: make this once a day

classes/task/delete_archived_object_metadata.php
keevan: image

I've added this under general settings, please let me know if the wording needs to be updated.

keevan:
Moodle 3.9.11+ (Build: 20211203), 567cf86dad927cde531f9ba9a84fea40cb61d2a9
Php: 7.3.28, pgsql: 10.17 (Debian 10.17-1.pgdg90+1), OS: Linux 5.4.0-91-generic x86_64
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

...............................................................  63 / 211 ( 29%)
.............................S.SSSSSSSSSSSSSSSSSS.SSSSSSSSSSSSS 126 / 211 ( 59%)
SSSSSSSSS.........SSS.......................................S.. 189 / 211 ( 89%)
......................                                          211 / 211 (100%)

Time: 19.71 seconds, Memory: 636.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 211, Assertions: 274, Skipped: 45.

Issue #457, opened 2021-12-21
De-duplicate signed urls to the same file based on rounded expiry dates


brendanheywood:
Lets say we have a course page which has poorly uploaded the same image 5 times, eg a 'course information' icon. A proper fix for this is being done elsewhere but this will help for some cases which get missed.

When someone views the course they will get 5 requests for the icon, each of which will serve a redirect to a signed url for that image in cloudfront. If those urls are slightly different then they will all be loading duplicates for the same image. So we should make some effort to make sure they are canonicalized as much as possible.

So in the case where the actual file name is identical, the main part which differs is the expiry. Due to the timing issues the expiry embedded in the url could be a second apart. So lets round the expiry to the nearest minute which means on average these should ave a muc higher collision rate and only a single real image will be loaded from cloudfront and cached in the browser.

brendanheywood:
Also make sure we always round down and not up

https://github.com/catalyst/moodle-tool_objectfs/blob/35951fde56b482f8c1bdc8e1dfd0b3ddd29e4317/classes/local/store/s3/client.php#L569-L594

PR #458, opened 2021-12-22
Round down s3 expiry to nearest minute, add more comments #457

keevan:
- This should mean any identical files that expiry within the same minute of each other, should theoretically now have the same URL and can be cached and prevent multiple loads from the browser.
- Also pushed out previous handling of expiry less than 1 minute to 2 minutes, which will be rounded down to the nearest minute by the new code.

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -571,26 +571,32 @@ class client extends object_client_base {
             // Convert to a valid timestamp.
             $expires = strtotime($expires);
         }
+        // Invalid or already expired:
         // If it's set to 0 or strtotime() returned false,
         // set it to default + 1 min as a healthy margin.
         if (empty($expires)) {
             $expires = $now + $this->expirationtime + MINSECS;
         }
-        // If it's already expired or expires less than in a minute,
-        // set it to 1 minute.
-        if ($expires < $now + MINSECS) {
-            $expires = $now + MINSECS;
+        // Expiry too short, push it out to the next 2 minutes (will round down later):
+        // If it's already expired or expires less than 2 minutes, set it to 2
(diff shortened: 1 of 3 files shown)


classes/local/store/s3/client.php
```diff
@@ -571,26 +571,32 @@ public function get_expiration_time($now, $expires) {
...
+            $expires = $now + (2 * MINSECS);
         }
+        // Checks upper bound:
         // The expiration date of a signature version 4 presigned URL must be
         // less than one week. So if it's greater than a week set it to 1 week.
         // Use MINSECS as a healthy margin of error.
         if ($expires - $now > WEEKSECS - MINSECS) {
             $expires = $now + WEEKSECS - MINSECS;
         }
-        if (is_null($expires) || false === $expires) {
```
keevan, line 590: This block would never have been reached, and is already covered by checks in lines 576-578 (prev commit)

classes/local/store/s3/client.php
```diff
@@ -571,26 +571,32 @@ public function get_expiration_time($now, $expires) {
...
-            // Invalid date format use default instead.
-            $expires = $now + $this->expirationtime;
-        }
+        // Rounds (down) to nearest minute:
+        // With our new expiry time, ensure we round down to the nearest minute
+        // (#457) to ensure expiry of potentially the same file will use the
+        // same URL, and will result in less duplicate requests.
+        $expires -= ($expires % MINSECS);
```
brendanheywood, line 598: All the logic looks good, can you amend the data provider here:

https://github.com/catalyst/moodle-tool_objectfs/blob/a43206024972ef1e61413e620971e90808844922/tests/object_file_system_test.php#L714

brendanheywood:
Good to go when green

Issue #368, opened 2020-07-28
Re normalize cut and pasted signed urls back into moodle plugin files


brendanheywood:
If someone views a PDF or file of some sort and then cut and pastes that url into a form and saves it then its now broken.

There are a couple entry points, but the vast majority will be processed by a formslib form in some way. There is a single touch point for processing this text before it is saved here:

https://github.com/moodle/moodle/blob/master/lib/filelib.php#L459-L476

I suggest we add a new callback point in that file and then object can implement it.  Similar stuff might need to be done in draftfiles too

https://github.com/moodle/moodle/blob/master/lib/filelib.php#L1266-L1278

There may also be a completely independent touch point in the repository api somewhere.

Examples:

Original url:

/pluginfile.php/1/tool_objectfs/settings/0/testimage.png

Becomes a signed url

https://abcdef.cloudfront.net/b5/f1/b5f150e80dd7f1d534f6da6d5c28d83ba665ac19?response-content-disposition=inline%3Bfilename%3D%22testimage.png%22&response-content-type=image%2Fpng&Expires=1595946738&Signature=........&Key-Pair-Id=ABCD

This must be 100% reversible into the original url. At the moment this url only contains the content hash and the filename but not the full relative path. This should be as easy as adding this in as an extra query param which is ignored by cloudfront

brendanheywood:
Doh:
> Signed CloudFront URLs cannot contain extra query string arguments. If you add a query string to a signed URL after you create it, the URL returns an HTTP 403 status.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html#private-content-how-signed-urls-work

This might still be viable but adding the extra roundtrip info into the url using a hash anchor instead of query params. This is likely more brittle but better than nothing.

brendanheywood:
We've tested this, it is viable after all, should not be that hard

PR #459, opened 2022-01-13
object_file_system: embed_origin_url(), rewrite_pluginfile_urls()

srdjan-catalyst:
#368 

redirect_to_presigned_url(): embed origin url into presigned url,
and corresponding reversal hook.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -47,6 +47,9 @@ require_once($CFG->libdir . '/filestorage/file_storage.php');
 
 abstract class object_file_system extends \file_system_filedir {
 
+    const PLUGINFILE_ORIGIN_SEGMENT_PREFIX = 'objectfs-origin=';
+
+    public $objectfs = true;
     public $externalclient;
     private $preferexternal;
     private $deleteexternally;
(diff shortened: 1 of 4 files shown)


classes/local/store/s3/client.php
```diff
@@ -422,6 +428,21 @@ public function support_presigned_urls() {
...
+     *
+     * @return string
+     */
+    public function presigned_domain() {
+        if ('cf' === $this->signingmethod) {
+            return $this->cloudfrontresourcedomain;
+        }
+
+        $key = 'fake-key';
+        $domain = $this->client->getObjectUrl($this->bucket, $key);
```
brendanheywood, line 442: does this call have any performance impact? ie does it do any network stuff or is it purely client side?

tests/object_file_system_test.php
```diff
@@ -930,4 +930,72 @@ public function test_is_configured_alternative_file_system_class_is_not_set() {
...
+        $this->assertEquals(
+            sprintf($textformat, $CFG->wwwroot.$ME),
+            $this->filesystem->rewrite_pluginfile_urls($originaltext)
+        );
+    }
+
+    /**
+     * Test rewrite_pluginfile_urls() CloudFront
+     */
+    public function test_rewrite_pluginfile_urls_s3_cf() {
```
brendanheywood, line 972: can you please turn this into a test with a dataProvider so we can throw a bunch of chunks of text at it and assert them all as a batch

classes/local/store/s3/client.php
srdjan-catalyst: Documentation says "Builds an URL to access object. Does not check if object exists, ACL etc". Maybe not exact words. So I take it does not do any S3 requests. I'll check anyway.

classes/local/store/object_file_system.php
```diff
@@ -804,13 +806,61 @@ protected function copy_from_external_to_local($contenthash) {
...
+            $presignedurl,
+            self::PLUGINFILE_ORIGIN_SEGMENT_PREFIX,
+            urlencode($me)
+        );
+    }
+
+    /**
+     * Convert redirected URLs in $text.
+     *
+     * @param string  $text The content that may contain ULRs in need of rewriting.
```
brendanheywood, line 842: typo ULR vs URL

classes/local/store/object_file_system.php
```diff
@@ -804,13 +806,61 @@ protected function copy_from_external_to_local($contenthash) {
...
+     * @return string The processed text.
+     */
+    public function rewrite_pluginfile_urls($text): string {
+        if (!$this->presigned_url_configured()) { // Do nothing.
+            return $text;
+        }
+
+        $re = sprintf(
+            '!%s\S*?\#%s(\S+\w)!',
+            preg_quote($this->externalclient->presigned_domain()),
```
brendanheywood, line 852: I'm just wondering if we can be more relaxed here and accept anything that looks like a url and has a fragment in the right shape. It would then work in a variety of edge cases like cutting and pasting urls between environments which either do or do no have signing on, or do have signing on with different domains. In all cases their local moodle urls should all have the same local part and Just Work (tm)

srdjan-catalyst: I actually started like that, but then thought we may want to limit the conversion. If we want to relax it then we need to embed full urls, not just paths to be glued to wwwroot. I suppose we can do that, and let acl decides what's available.

bwalkerl:
Closing in favour of #758 which uses this as a base and makes use of a new core hook.

Issue #212, opened 2020-01-18
Add option to manage trashdir remotely


brendanheywood:
Probably need to be aware of whether the remote object store is actually doing remote deletes or not

Settings:
* [ ] manage trashdir on disk
* [ ] manage trashdir virtually with a separate table of trashed file and then a final delete after x days
* [ ] no trashdir at all, deletes are instant (or not at all if a local policy of deletes are soft deletes).

Note if prod is set to not delete objects ever, then effectively trashdir is completely pointless. So if we have a setting which informs objectfs that things are kept we can simply not copy them to trashdir in the first place.

brendanheywood:
This is sorted in https://github.com/catalyst/moodle-tool_objectfs/pull/429

brendanheywood:
Reopening as looks like this is only half done

brendanheywood:
Internal comms:

so here is my rough thinking, at a high level and completely ignoring the current implementation in core, trashdir is when you've deleted the last mdl_files reference to a file, we do some special handling to enable a file to be recovered manually even if you cannot recover the mdl_files record itself. With objectfs with store deletes turned off you get this for free as all files are in the store forver, all you need is their hash and you can go grab it. We always know the hash because we keep it in tool_objectfs_objects with a state of ORPHANED. So we have all the metadata to fully implement a proper trashdir done the objectfs way without any extra effort other that removing all the crap that we inherit from core and don't want. So when we delete the last reference we should not do any copying of the file from the store to anywhere, not local, not to another store bucket, it just stays where it is. The metadata is already sorted. 

task/delete_orphaned_object_metadata.php will already optionally delete the metadata after some time which means we could still recover the file if we knew its hash but we won't have an easy record of it. But for instance if someone uploads an old course backup which doesn't contain files, but it does have the hash then it will auto heal and the file will magically appear again (assuming we don't delete)

if we want a true trashdir function then that task can be augmented or another task made which not only deletes the metadata but also does the store file delete as well

no directory scans are ever needed at any point, trashdir in datadir will never exist

and no file renaming or prefixes

danmarsden:
@brendanheywood - is this patch (untested) along the lines of what you were thinking?
https://github.com/catalyst/moodle-tool_objectfs/compare/trashhandling

thanks!

brendanheywood:
Yup the direction here looks good

more bits we should be able to rip out:
1) override the file_system constructor so we never test for trashdir's existance nor auto create it
[...]

PR #465, opened 2022-02-14
Tidy up trash handling process for file deletes.

danmarsden:
This removes the semi-broken "move to trash location" handling of objectfs and relies on the cleanup task to delete files if set.

One challenge we have with this is that if a site starts with "don't delete" and then moves to "delete using the orphaned file" process is that it won't really delete orphans - only orphans that Moodle still knows about.

Note - this is currently untested live - we should hold off merging until we test this properly, but comments/feedback on the approach prior to then is welcome.

### DIFF
--- a/classes/task/delete_orphaned_object_metadata.php
+++ b/classes/task/delete_orphaned_object_metadata.php
@@ -15,8 +15,8 @@
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
 /**
- * Task that checks for old orphaned objects, and removes their metadata
- * (record) as it is no longer useful/relevant.
+ * Task that checks for old orphaned objects, and removes their metadata (record)
+ * and external file (if delete external enabled) as it is no longer useful/relevant.
  *
  * @package   tool_objectfs
  * @author    Kevin Pham <kevinpham@catalyst-au.net>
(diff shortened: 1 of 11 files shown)


danmarsden:
@dmitriim @brendanheywood - keen on your thoughts for this approach - more info in #212

dmitriim:
pinging @golenkovm

classes/task/delete_orphaned_object_metadata.php
```diff
@@ -52,10 +52,28 @@ public function execute() {
...
         ];
-        $count = $DB->count_records_select('tool_objectfs_objects', $wheresql, $params);
-        if (!empty($count)) {
-            mtrace("Deleting $count records with orphaned metadata (orphaned tool_objectfs_objects)");
-            $DB->delete_records_select('tool_objectfs_objects', $wheresql, $params);
+
+        if (!empty($this->config->deleteexternal) && $this->config->deleteexternal == TOOL_OBJECTFS_DELETE_EXTERNAL_TRASH) {
+            // We need to delete the external files as well as the orphaned data.
+            $filesystem = new $this->config->filesystem();
+            $objects = $DB->get_recordset_select('tool_objectfs_objects', $wheresql, $params);
```
danmarsden, line 59: I'm tempted to sanity check this by using some sql that will only return tool_objectfs_objects where the contenthash doesn't exist in the files table - @brendanheywood that make sense to you?

danmarsden:
@brendanheywood - I've tested this patch on a local site and the deletion of orphaned files works well - I'm ready for this to be PR'd and merged, although I'm still interested to know if you think I should sanity check the sql the orphaned data clean up so that it will only return tool_objectfs_objects records where the contenthash doesn't exist in the files table.

PR #468, opened 2022-03-03
ObjectFS: Send correct expiry time for redirect URLs #422

sammarshallou:
When redirecting the user to download a file directly from S3, the
redirect was previously sent with whatever cache expiry time that
Moodle thought was appropriate for the file. In cases where this was
longer than a week, it caused problems, because S3 signed URLs can
only be valid for up to a week. This change sets the expiry on the
redirect to the same time that the URL expires at.

#468 #422

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -736,8 +736,29 @@ abstract class object_file_system extends \file_system_filedir {
      * @throws \dml_exception
      */
     public function redirect_to_presigned_url($contenthash, $headers = array()) {
+        global $FULLME;
         try {
-            redirect($this->externalclient->generate_presigned_url($contenthash, $headers));
+            $signedurl = $this->externalclient->generate_presigned_url($contenthash, $headers);
+            if (headers_sent()) {
+                debugging('objectfs redirect for ' . $contenthash . ' from ' . $FULLME .
+                        ': headers already sent; redirect may be incorrectly cached in browser');
+            } else {
+                // Remove all previously-set headers.
+                foreach (headers_list() as $header) {
+                    // Get header name (text before the colon).
+                    if (preg_match('~^([^:]+):~', $header, $matches)) {
+                        header_remove($matches[1]);
+                    }
+                }
+                // Set expires and cache-control values to match the presigned URL expiry, which may be
+                // different to values previously set.
+                header('Expires: '. gmdate('D, d M Y H:i:s', $signedurl->expiresat) .' GMT');
+                // Use 'private' to allow browser caching only, otherwise via a shared cache users might
+                // be able to redirect to content that was only supposed to be displayed to a different
+                // user.
+                header('Cache-Control: private, max-age=' . ($signedurl->expiresat - time()));
+            }
+            redirect($signedurl->url);
         } catch (\Exception $e) {
             debugging('Failed to redirect to pre-signed url: ' . $e->getMessage());
             return false;
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -743,20 +743,30 @@ abstract class object_file_system extends \file_system_filedir {
                 debugging('objectfs redirect for ' . $contenthash . ' from ' . $FULLME .
                         ': headers already sent; redirect may be incorrectly cached in browser');
             } else {
-                // Remove all previously-set headers.
+                // Remove all previously-set headers, and look for cache-control setting.
+                $cachecontrol = '';
                 foreach (headers_list() as $header) {
                     // Get header name (text before the colon).
-                    if (preg_match('~^([^:]+):~', $header, $matches)) {
+                    if (preg_match('~^([^:]+):(.*)$~', $header, $matches)) {
+                        if (strtolower($matches[1]) === 'cache-control') {
+                            $cachecontrol = $matches[2];
+                        }
                         header_remove($matches[1]);
                     }
                 }
                 // Set expires and cache-control values to match the presigned URL expiry, which may be
                 // different to values previously set.
                 header('Expires: '. gmdate('D, d M Y H:i:s', $signedurl->expiresat) .' GMT');
-                // Use 'private' to allow browser caching only, otherwise via a shared cache users might
+                // Unless cache-control was previously set to 'public' by Moodle for the actual file send,
+                // use 'private' to allow browser caching only; otherwise via a shared cache users might
                 // be able to redirect to content that was only supposed to be displayed to a different
                 // user.
-                header('Cache-Control: private, max-age=' . ($signedurl->expiresat - time()));
+                if (strpos($cachecontrol, 'public') !== false) {
+                    $publicprivate = 'public';
+                } else {
+                    $publicprivate = 'private';
+                }
+                header('Cache-Control: ' . $publicprivate . ', max-age=' . ($signedurl->expiresat - time()));
             }
             redirect($signedurl->url);
         } catch (\Exception $e) {
--- a/classes/local/store/signed_url.php
+++ b/classes/local/store/signed_url.php
@@ -1,0 +1,46 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


classes/local/store/object_file_system.php
```diff
@@ -737,7 +737,15 @@ protected function copy_from_external_to_local($contenthash) {
...
         try {
-            redirect($this->externalclient->generate_presigned_url($contenthash, $headers));
+            $signedurl = $this->externalclient->generate_presigned_url($contenthash, $headers);
+            // Set expires and cache-control values to match the presigned URL expiry, which may be
+            // different to values previously set.
+            header('Expires: '. gmdate('D, d M Y H:i:s', $signedurl->expiresat) .' GMT');
+            // Use 'private' to allow browser caching only, otherwise via a shared cache users might
+            // be able to redirect to content that was only supposed to be displayed to a different
+            // user.
+            header('Cache-Control: private, max-age=' . ($signedurl->expiresat - time()));
```
brendanheywood, line 747: This all looks great @sammarshallou 

There is still a part of me that wants to go the one extra step and do the full nuclear option where we completely unset all headers that have been set but not flushed by _pluginfile so that we only ever get the exact headers we do here. There are numerous ways of setting expires and it's really unclear to me how different browser might priorities conflicting headers.

ie loop over headers_list() and unset them. If headers_sent() already then we have to keep going but may log a debugging call

sammarshallou: OK! I have force-pushed a modified commit that uses header_remove to get rid of existing headers, and also adds a debugging message if headers were already sent. It took me 2 attempts because my first one used header_remove(null) but then I spotted the bit at the bottom of the manual page that says that only works in PHP 8, so I've changed it to loop around as suggested.

The only changes to previous commit are in that same section quoted above.

classes/local/store/object_file_system.php
```diff
@@ -737,7 +737,26 @@ protected function copy_from_external_to_local($contenthash) {
      */
     public function redirect_to_presigned_url($contenthash, $headers = array()) {
         try {
-            redirect($this->externalclient->generate_presigned_url($contenthash, $headers));
+            $signedurl = $this->externalclient->generate_presigned_url($contenthash, $headers);
+            if (headers_sent()) {
+                debugging('Headers already sent! This redirect may be incorrectly cached in browser');
```
brendanheywood, line 742: Can you just stuff this message with some useful context, mention objectfs and the url we are serving

classes/local/store/object_file_system.php
```diff
@@ -737,7 +737,26 @@ protected function copy_from_external_to_local($contenthash) {
...
+            } else {
+                // Remove all previously-set headers.
+                foreach (headers_list() as $header) {
+                    // Get header name (text before the colon).
+                    if (preg_match('~^([^:]+):~', $header, $matches)) {
+                        header_remove($matches[1]);
+                    }
+                }
```
keevan, line 750: This might be more simply done as such:
```
 header_remove(null);
```
Ref: https://www.php.net/manual/en/function.header-remove.php

classes/local/store/signed_url.php
```diff
@@ -0,0 +1,50 @@
...
+ * This can be obtained using the object_client::generate_presigned_url function.
+ *
+ * @package tool_objectfs
+ * @copyright 2022 The Open University
+ * @license http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class signed_url {
+    /**
+     * @var \moodle_url URL to redirect to
+     */
```
keevan, line 34: Can be a one line comment instead
```suggestion
    /** @var \moodle_url URL to redirect to */
```

classes/local/store/signed_url.php
```diff
@@ -0,0 +1,50 @@
...
+    public $url;
+
+    /**
+     * @var int Expiry timestamp (Unix epoch) after which this URL will stop working
+     */
```
keevan, line 39: Similarly with the above.
```suggestion
    /** @var int Expiry timestamp (Unix epoch) after which this URL will stop working */
```

keevan (commented):
Everything else seems okay

classes/local/store/object_file_system.php
```diff
@@ -737,7 +737,26 @@ protected function copy_from_external_to_local($contenthash) {
...
+                // Set expires and cache-control values to match the presigned URL expiry, which may be
+                // different to values previously set.
+                header('Expires: '. gmdate('D, d M Y H:i:s', $signedurl->expiresat) .' GMT');
+                // Use 'private' to allow browser caching only, otherwise via a shared cache users might
+                // be able to redirect to content that was only supposed to be displayed to a different
+                // user.
+                header('Cache-Control: private, max-age=' . ($signedurl->expiresat - time()));
```
brendanheywood, line 757: Adding Cache-Control: private I *think* should be safe, but not optimal, in all cases, I've written up a new feature here but I am pretty sure it's not a blocker for this patch but would like some eyes to double check:

https://github.com/catalyst/moodle-tool_objectfs/issues/469

classes/local/store/object_file_system.php
brendanheywood: See Sam's previous comment

classes/local/store/signed_url.php
sammarshallou: Thank you, yes that's neater, fixed.

classes/local/store/object_file_system.php
sammarshallou: Done, I added the contenthash as well (probably not super useful but just in case).

classes/local/store/object_file_system.php
sammarshallou: Hmm interesting - so you are suggesting that we could look at the existing cache-control header value and if there is one which already includes 'public' then just use public instead of private here? I hadn't thought of that. I think we could do that here in this patch and then don't need the separate issue, it doesn't sound too risky (if there is a case where Moodle sets cache-control public for these URLs already when they shouldn't be public, then that would already be a bug in the same way, so hopefully it doesn't). I'll have a quick look at that and then comment again when done (writing this now just in case I get distracted before I can finish it).

sammarshallou: I've pushed a change to allow 'public' as a second commit on this branch (so we can squash it into the original commit if you like, or get rid of it if it's not a good idea). Obviously this doesn't do the full thing with actually using unsigned URLs that you were proposing, only makes the redirect caching public.

If I've understood it correctly, I don't think there is a security hole in leaking a URL that was signed for user A so that user B can see it, except that it allows user B to access that single file, which in this case is what we want to happen since Moodle was serving the file with cache-control public.

If we send it 'public' now then in previous behaviour before this change (ie before the code to remove all the headers) we would have been sending it with cache-control public already so it at least shouldn't worsen security...

classes/local/store/object_file_system.php
```diff
@@ -736,8 +736,39 @@ protected function copy_from_external_to_local($contenthash) {
...
+            } else {
+                // Remove all previously-set headers, and look for cache-control setting.
+                $cachecontrol = '';
+                foreach (headers_list() as $header) {
+                    // Get header name (text before the colon).
+                    if (preg_match('~^([^:]+):(.*)$~', $header, $matches)) {
+                        if (strtolower($matches[1]) === 'cache-control') {
+                            $cachecontrol = $matches[2];
+                        }
+                        header_remove($matches[1]);
```
keevan, line 754: It would be nicer if `$matches[1]` and `$matches[2]` were named variables which describes what they are, e.g. `name` and `value` (or `headername` and `headervalue`).

This would likely make it quicker to read and reason about.

classes/local/store/object_file_system.php
```diff
@@ -736,8 +736,39 @@ protected function copy_from_external_to_local($contenthash) {
...
+                header('Expires: '. gmdate('D, d M Y H:i:s', $signedurl->expiresat) .' GMT');
+                // Unless cache-control was previously set to 'public' by Moodle for the actual file send,
+                // use 'private' to allow browser caching only; otherwise via a shared cache users might
+                // be able to redirect to content that was only supposed to be displayed to a different
+                // user.
+                if (strpos($cachecontrol, 'public') !== false) {
+                    $publicprivate = 'public';
+                } else {
+                    $publicprivate = 'private';
+                }
```
keevan, line 768:
```suggestion
                $publicprivate = 'private';
                if (strpos($cachecontrol, 'public') !== false) {
                    $publicprivate = 'public';
                }
```
I prefer less `else` blocks (reduces mental overhead). This would achieve more or less the same output and reads better personally.

keevan: Also could you change the variable name to `$cachevisibility` instead?

brendanheywood:
thanks heaps @sammarshallou as always :)

sammarshallou:
Thanks all :)

Issue #472, opened 2022-03-23
Exception - The HeadBucket operation requires non-empty parameter: Bucket


dmitriim:
- Navigate to /admin/settings.php?section=tool_objectfs_settings and set tool_objectfs | filesystem to \tool_objectfs\s3_file_system

You get following error and the setting page is broken. So the only way to recover is `php admin/cli/cfg.php --component=tool_objectfs --name=filesystem --unset`

```
Exception - The HeadBucket operation requires non-empty parameter: Bucket

[More information about this error](https://docs.moodle.org/39/en/error/moodle/generalexceptionmessage)

×Debug info:
Error code: generalexceptionmessage
×Stack trace:
line 64 of /local/aws/sdk/Aws/InputValidationMiddleware.php: InvalidArgumentException thrown
line 80 of /local/aws/sdk/Aws/Middleware.php: call to Aws\InputValidationMiddleware->__invoke()
line 578 of /local/aws/sdk/Aws/S3/S3Client.php: call to Aws\Middleware::Aws\{closure}()
line 601 of /local/aws/sdk/Aws/S3/S3Client.php: call to Aws\S3\S3Client::Aws\S3\{closure}()
line 535 of /local/aws/sdk/Aws/S3/S3Client.php: call to Aws\S3\S3Client::Aws\S3\{closure}()
line 554 of /local/aws/sdk/Aws/S3/S3Client.php: call to Aws\S3\S3Client::Aws\S3\{closure}()
line 54 of /local/aws/sdk/Aws/Middleware.php: call to Aws\S3\S3Client::Aws\S3\{closure}()
line 59 of /local/aws/sdk/Aws/S3/SSECMiddleware.php: call to Aws\Middleware::Aws\{closure}()
line 77 of /local/aws/sdk/Aws/IdempotencyTokenMiddleware.php: call to Aws\S3\SSECMiddleware->__invoke()
line 64 of /local/aws/sdk/Aws/AwsClientTrait.php: call to Aws\IdempotencyTokenMiddleware->__invoke()
line 58 of /local/aws/sdk/Aws/AwsClientTrait.php: call to Aws\AwsClient->executeAsync()
line 86 of /local/aws/sdk/Aws/AwsClientTrait.php: call to Aws\AwsClient->execute()
line 210 of /admin/tool/objectfs/classes/local/store/s3/client.php: call to Aws\AwsClient->__call()
[...]

Issue #467, opened 2022-03-01
Skip connection test for certain storage file systems (e.g. s3) if certain required fields aren't set


keevan:
In the case of S3, this should reduce/remove the initial load time filled by the connection test, and alllow you to enter in the relevant details.

The keys to check and skip the connection test should likely be (at least for s3):
- s3_key
- s3_secret
- s3_bucket
- s3_region
- s3_base_url (maybe)

I'm not 100% sure about the other fields, but these seem like mandatory fields that if aren't set, would cause the connection to fail anyways, so it _should fail immediately_.

There is also the possibility of improving the wait period of this by setting a timeout on the connection test. 
More info: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_configuration.html#http-connect-timeout

dmitriim:
will help to resolve https://github.com/catalyst/moodle-tool_objectfs/issues/472

dmitriim:
should we count this one as done? @keevan

keevan:
I think a configurable / sane default timeout might also be useful here, for example, if it was configured but maybe the configuration was not 100% correct.

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_configuration.html#http-connect-timeout

This might be a bonus though, but I'll leave it there for discussion.

PR #473, opened 2022-03-23
issue #472: check if s3 client configured before using it

dmitriim:


classes/local/store/s3/client.php
```diff
@@ -79,44 +79,91 @@ public function __wakeup() {
...
         }
 
-        // Support base_url config for aws api compatible endpoints.
-        if ($config->s3_base_url) {
-            $options['endpoint'] = $config->s3_base_url;
+        if (empty($config->s3_usesdkcreds)) {
+            if (empty($config->s3_key) && empty($config->s3_secret)) {
+                return false;
+            }
         }
```
keevan, line 114: I think this was meant to be something like this according to https://github.com/catalyst/moodle-tool_objectfs/issues/467

```suggestion
        if (empty($config->s3_usesdkcreds)) {
            return false;
        }
        
        if (empty($config->s3_key) && empty($config->s3_secret)) {
            return false;
        }
```

classes/local/store/s3/client.php
```diff
@@ -79,44 +79,91 @@ public function __wakeup() {
...
+            if ($config->useproxy) {
+                $options['http'] = array('proxy' => $this->get_proxy_string());
+            }
+
+            // Support base_url config for aws api compatible endpoints.
+            if ($config->s3_base_url) {
+                $options['endpoint'] = $config->s3_base_url;
+            }
+
+            $this->client = \Aws\S3\S3Client::factory($options);
```
keevan, line 146: I believe it would read easier and be easier to maintain long term using guards.

```suggestion
        // Not configured.
        if (!$this->is_configured($config)) {
            $this->client = null;
            return;
        } 
        
        $options = array(
            'region' => $config->s3_region,
            'version' => AWS_API_VERSION
        );

        if (empty($config->s3_usesdkcreds)) {
            $options['credentials'] = array('key' => $config->s3_key, 'secret' => $config->s3_secret);
        }

        if ($config->useproxy) {
            $options['http'] = array('proxy' => $this->get_proxy_string());
        }

        // Support base_url config for aws api compatible endpoints.
        if ($config->s3_base_url) {
            $options['endpoint'] = $config->s3_base_url;
        }

        $this->client = \Aws\S3\S3Client::factory($options);
```

classes/local/store/s3/client.php
```diff
@@ -79,44 +79,91 @@ public function __wakeup() {
...
-        $options = array(
-            'region' => $config->s3_region,
-            'version' => AWS_API_VERSION
-        );
+    /**
+     * Check if the client is functional.
+     * @return bool
+     */
+    private function is_functional() {
+        return !is_null($this->client);
```
keevan, line 92: Saves an op. That and isset being a language construct it should be slightly faster than a function call.

```suggestion
        return isset($this->client);
```

classes/local/store/s3/client.php
```diff
@@ -79,44 +79,91 @@ public function __wakeup() {
...
+        if (!$this->is_configured($config)) {
+            $this->client = null;
+        } else {
+            $options = array(
+                'region' => $config->s3_region,
+                'version' => AWS_API_VERSION
+            );
+
+            if (empty($config->s3_usesdkcreds)) {
+                $options['credentials'] = array('key' => $config->s3_key, 'secret' => $config->s3_secret);
```
keevan, line 134: I prefer square brackets for arrays, but this being on multiple lines makes more sense imo. And multi line arrays are better using the square bracket form because of the ability to use a trailing comma without potential compatibility disaster :smile: 
```suggestion
                $options['credentials'] = [
                    'key' => $config->s3_key,  
                    'secret' => $config->s3_secret,
                ];
```

classes/local/store/s3/client.php
```diff
@@ -79,44 +79,91 @@ public function __wakeup() {
...
     public function register_stream_wrapper() {
-        if ($this->get_availability()) {
+        if ($this->get_availability() && $this->is_functional()) {
             $this->client->registerStreamWrapper();
         } else {
             parent::register_stream_wrapper();
         }
     }
 
     private function get_md5_from_hash($contenthash) {
```
keevan, line 162: Can you please add phpdocs to this function (In particular params and return type), I see false is already expected e.g. from an exception throw but just to be clearer upfront.

classes/local/store/s3/client.php
```diff
@@ -206,17 +253,23 @@ public function test_connection() {
...
             $connection->success = false;
-            $connection->details = $this->get_exception_details($e);
-        } catch (\GuzzleHttp\Exception\InvalidArgumentException $e) {
-            $connection->success = false;
-            $connection->details = $this->get_exception_details($e);
-        } catch (\Aws\Exception\CredentialsException $e) {
-            $connection->success = false;
-            $connection->details = $this->get_exception_details($e);
+            $connection->details = '';
+        } else {
```
keevan, line 259: I would recommend de-nesting this block, e.g. one if statement with an early return. Everything else should flow as it would in the _ideal case_, matching the function's name/action, which is to test the connection.

I also think the `$connection->details = '';` which be a get_string probably something along the lines of `the s3 client is not functional, possibly because it has not been set up or configured yet`

classes/local/store/s3/client.php
```diff
@@ -235,49 +288,54 @@ public function test_permissions($testdelete) {
...
-        try {
-            $result = $this->client->getObject(array(
-                            'Bucket' => $this->bucket,
-                            'Key' => $this->bucketkeyprefix . 'permissions_check_file'));
-        } catch (\Aws\S3\Exception\S3Exception $e) {
-            $errorcode = $e->getAwsErrorCode();
-            // Write could have failed.
-            if ($errorcode !== 'NoSuchKey') {
+            $permissions->messages = array();
+        } else {
```
keevan, line 294: Can we de-nest this block as well please? Same reasons as prior.

classes/local/store/s3/client.php
```diff
@@ -235,49 +288,54 @@ public function test_permissions($testdelete) {
...
-                $permissions->messages[get_string('settings:deletesuccess', 'tool_objectfs')] = 'warning';
-                $permissions->success = false;
+                $result = $this->client->getObject(array(
+                    'Bucket' => $this->bucket,
+                    'Key' => $this->bucketkeyprefix . 'permissions_check_file'));
             } catch (\Aws\S3\Exception\S3Exception $e) {
                 $errorcode = $e->getAwsErrorCode();
-                // Something else went wrong.
-                if ($errorcode !== 'AccessDenied') {
+                // Write could have failed.
```
keevan, line 312: This comment seems misplaced (maybe it belongs in the previous block for `writefailure`

This is indicating a read failure I believe.

keevan (changes requested):
Various styling cleanup. Otherwise, I think the logic added is good.

keevan:
It probably wouldn't hurt to also add a simple test to ensure it doesn't explode, e.g. if the connection was not set up fully.

classes/local/store/s3/client.php
dmitriim: we want to check s3_key and s3_secret ONLY if s3_usesdkcreds is emptyt, don't we?

classes/local/store/s3/client.php
dmitriim: I'm not a fan of doing that. But I can see that it's been done every where in the code. So changing it...

classes/local/store/s3/client.php
dmitriim: yeah agree. isset would be better here. changing...

classes/local/store/s3/client.php
dmitriim: 100% agree. But, formatting should be fixed in a different PR.

classes/local/store/s3/client.php
keevan: Ah right. I get what you mean. I haven't pulled it back up but assuming `s3_usesdkcreds` is on the same lines as "Use the environment to determine the keys available" (to a certain extent), I think maybe this is the preferred form - i.e. don't nest the conditions, just add to it since it's all AND checks.

```suggestion
        if (empty($config->s3_usesdkcreds)
            && empty($config->s3_key) 
            && empty($config->s3_secret)) {
            return false;
        }
```

classes/local/store/s3/client.php
dmitriim: good point about an extra string to explain the failure. But don't think returning earlier in this case would make the code better. However, I have updated the code and moved if block inside try catch.

classes/local/store/s3/client.php
dmitriim: removing completely.

dmitriim:
any idea on where to add this test?

classes/local/store/s3/client.php
dmitriim: I wrote it in a way a think about it as I wasn't sure if there could be any other combinations in case if s3_usesdkcreds is set. As there is only one at the moment, I'm changing to a suggested code.

classes/local/store/s3/client.php
dmitriim: oh. Ideally this method should be refactored... Applying de-nesting...

classes/local/store/s3/client.php
dmitriim: should be in different MR

classes/local/store/s3/client.php
dmitriim: done

classes/local/store/s3/client.php
dmitriim: done

classes/local/store/s3/client.php
dmitriim: removed the comment

classes/local/store/s3/client.php
```diff
@@ -235,11 +286,17 @@ public function test_permissions($testdelete) {
         $permissions->success = true;
         $permissions->messages = array();
 
+        if ($this->is_functional()) {
+            $permissions->success = false;
+            $permissions->messages = array();
+            return  $permissions;
```
keevan, line 292: Double spacing
```suggestion
            return $permissions;
```

keevan:
> any idea on where to add this test?

Perhaps here? `admin/tool/objectfs/tests/classes/test_s3_integration_client.php`

classes/local/store/s3/client.php
dmitriim: good catch. Fixed.

dmitriim:
added tests. They actually helped to find a bug in is_configured for s3

keevan (approved):

Issue #471, opened 2022-03-23
Coding error detected, it must be fixed by a programmer: Objectfs does not implement a trashdir.


dmitriim:
On a freshly upgraded from 3.5 to 3.9 instance (site-data is potentially missing). Object fs is set as a file system, but not configured. 

```
Coding error detected, it must be fixed by a programmer: Objectfs does not implement a trashdir.

More information about this error

×Debug info:
Error code: codingerror
×Stack trace:
line 1022 of /admin/tool/objectfs/classes/local/store/object_file_system.php: coding_exception thrown
line 253 of /lib/filestorage/file_system_filedir.php: call to tool_objectfs\local\store\object_file_system->get_trash_fullpath_from_hash()
line 596 of /admin/tool/objectfs/classes/local/store/object_file_system.php: call to file_system_filedir->recover_file()
line 127 of /lib/filestorage/file_system_filedir.php: call to tool_objectfs\local\store\object_file_system->recover_file()
line 384 of /lib/filestorage/file_system.php: call to file_system_filedir->get_local_path_from_storedfile()
line 583 of /lib/filestorage/stored_file.php: call to file_system->get_imageinfo()
line 598 of /lib/filestorage/stored_file.php: call to stored_file->get_imageinfo()
line 1284 of /course/renderer.php: call to stored_file->is_valid_image()
line 1359 of /course/renderer.php: call to core_course_renderer->course_overview_files()
line 1223 of /course/renderer.php: call to core_course_renderer->coursecat_coursebox_content()
line 1446 of /course/renderer.php: call to core_course_renderer->coursecat_coursebox()
line 2153 of /course/renderer.php: call to core_course_renderer->coursecat_courses()
line 2452 of /course/renderer.php: call to core_course_renderer->frontpage_available_courses()
line 121 of /index.php: call to core_course_renderer->frontpage()
```

danmarsden:
Thanks @dmitriim - this is a regression from #465 

What happens here is that it tries to recover the missing file in object storage, which fails, so it  then calls the core file api recover file function,
https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/object_file_system.php#L596

problem is that the core recover file fucntion tries to access the external storage trashdir which now throws an exception - I'm guessing we need to stop throwing an exceptions in get_trash_fullpath_from_hash and just return "false" ? - not ideal, but at least that will prevent the fatal error.

dmitriim:
[...]

PR #476, opened 2022-03-24
iassue #471: do not throw exception on trash dir related methods

dmitriim:


brendanheywood:
-1 I'd prefer to keep this in place to highlight anymore errors we've potentially missed.

See https://github.com/catalyst/moodle-tool_objectfs/issues/471#issuecomment-1077177704

PR #477, opened 2022-04-16
Add missing "Principal": "*" in S3 bucket access policy JSON

nadavkav:
It seems AWS updated their S3 bucket access policy settings, which now requires "Principal": "*" in each statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,13 @@ There is support for more object stores planed.
   "Version": "2012-10-17",
   "Statement": [
     {
+      "Principal": "*",
       "Effect": "Allow",
       "Action": ["s3:ListBucket"],
       "Resource": ["arn:aws:s3:::bucketname"]
     },
     {
+      "Principal": "*",
       "Effect": "Allow",
       "Action": [
         "s3:PutObject",


brendanheywood:
thanks @nadavkav

brendanheywood:
Actually this appears to be a mistake and we'll roll it back:

> this is not a bucket policy, and if it gets used as one it gives the entire world unlimited read and write access to the bucket

It is also against the wrong branch, so right now won't do anything anyway

nadavkav:
@brendanheywood , what do you mean? did I used the wrong policy? should I configure it for something more restricting?

brendanheywood:
I'm not fully across this, just relaying what the inf team said. Some googling seems to confirm this:

https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html#example-bucket-policies-use-case-2

I'm not sure if you are using url signing, but if you are there are two ways that url signing can be set up, using s3 signing and using cloudfront signing. We use the latter and have not invested much time testing or refining the docs for pure s3 signing. We started down that road and quickly swapped. So maybe the policy for pure s3 looks different but I can't offer much advice on that front. I believe others in the wild are using that method.

nadavkav:
I am using exactly the example you pointed to at AWS.
And this is were I found that the policy template in the code is missing:
"Principal": "*"

I am not using  url signing
Should I limit that access using IP of my Moodle servers?
As they are getting the Moodle data files from S3, and then serve them via CF publicly,
Did I understand the architecture correctly?

keevan:
Hi @nadavkav, as described in the documentation @brendanheywood linked, you may use the `arn::**` in the Principal section to reference users and/or services that can access the bucket. 

This can be easier, but as shown, it is also easy to misconfigure it such that anyone could read/write to the buckets. The alternative here is to set up an IAM policy for a user / group, to have explicit access to the bucket. This is configuring access from the other side (user to bucket) instead of bucket for user(s). 

Reference: _Granting permissions to multiple accounts with added conditions_ [AWS Bucket policies](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html#example-bucket-policies-use-case-2)

PR #478, opened 2022-04-22
New file sync filter settings: mimetype, filearea, component

nadavkav:
A new feature, that allow the admin to filter what files are sync into the remote storage based on combination of the following Moodle mdl_files table fields: component, filearea, mimetype.
There is an AND logical condition between the above fields.


nadavkav:
@brendanheywood How about this?
We are using it for a few months now, and it seem to work fine.

brendanheywood:
Again this one is targeting a deprecated branch can you please rebase it. Also we have a decent refactor which is about to land to improve performance, we hope to land that this week we are currently load testing it:

https://github.com/catalyst/moodle-tool_objectfs/pull/506

I would assume this will clash with yours, so ideally if you can rebase on it on top of that one, but might be easier to hold off until it lands and then rework it

nadavkav:
I will wait until the new version lands, and then rebase all my various PRs

PR #482, opened 2022-05-02
Fix broken range requests

### LINKED ISSUE #481: Proxy range requests - Test range request failed
image

This may be related to recent changes in AWS SDK https://github.com/catalyst/moodle-local_aws/pull/36

golenkovm:
This is to fix #481

### DIFF
--- a/classes/check/proxy_range_request.php
+++ b/classes/check/proxy_range_request.php
@@ -1,0 +1,69 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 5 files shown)


classes/check/proxy_range_request.php
```diff
@@ -0,0 +1,69 @@
...
+        $url = new \moodle_url('/admin/category.php', ['category' => 'tool_objectfs']);
+        return new \action_link($url, get_string('pluginname', 'tool_objectfs'));
+    }
+
+    /**
+     * Check for the success of a proxied range request, if the setting is enabled.
+     *
+     * @return result
+     */
+    public function get_result() : result {
```
keevan, line 45:
```suggestion
    public function get_result(): result {
```

lib.php
```diff
@@ -110,3 +110,18 @@ function tool_objectfs_pluginfile($course, $cm, context $context, $filearea, arr
...
+ * @return array
+ */
+function tool_objectfs_status_checks() {
+    if (get_config('tool_objectfs', 'proxyrangerequests')) {
+        return [
+            new tool_objectfs\check\proxy_range_request()
+        ];
+    } else {
+        return [];
+    }
```
keevan, line 126:
```suggestion
    return [];
```

classes/check/proxy_range_request.php
Peterburnett: Thanks @keevan , fixed

golenkovm:
Cool, it's looking good @Peterburnett 
By the way, should we also have a check for objectfs connection status?

keevan (approved):

PR #483, opened 2022-05-10
Added use path endpoint setting for s3

danielannan:
Some S3 type services doesn't support access to bucket by subdomain, this adds a checkbox setting for configuring the s3 client to use the path for accessing the bucket on the endpoint.


John-Bob-DIRIENZO:
This PR is pretty useful ! Thanks.

PR #456, opened 2021-12-20
S3 pre-signed URL: error with unicode filename #455

PhucNguyen0311:
I have submitted a fix for issue "S3 pre-signed URL: error with unicode filename" #455. Please help me to review this.

Issue #455, opened 2021-12-20
S3 pre-signed URL: error with unicode filename 


PhucNguyen0311:
Hi, I got an issue relative to S3 storage with Unicode filename.
![image](https://user-images.githubusercontent.com/67891661/146698753-95742ab3-7b34-4637-a4e7-970b49d8fe03.png)

Currently, the CloudFront pre-signed method (generate_presigned_url_cloudfront) uses a function to get a nice filename but S3 isn't. So, I think it would be good if S3 pre-signed function (generate_presigned_url_s3) has the same method.

marxjohnson:
@PhucNguyen0311 This looks like the right approach to me, however it introduces some duplicated logic. As you say, the cloudfront and s3 signing methods should do the same thing, so it would be good if they both did this by calling `get_nice_filename()`. Perhaps you could refactor that function slightly so that it returns the data in a format that can be used by both functions? This may require a small change to the cloudfront method too. For example, `get_nice_filename()` could return an array of content-disposition, filename and content-type, then `generate_presigned_url_cloudfront()` could use that to construct its query string while `generate_presigned_url_s3()` could use to it to construct its parameters.

marxjohnson:
@PhucNguyen0311 This fix looks good to me, @brendanheywood what do you think?

brendanheywood:
@PhucNguyen0311 can you please rebase your patch

marxjohnson:
@brendanheywood I tried to contact @PhucNguyen0311 but didn't get a response, so I have done the rebase and created a new PR: #493

tuanngocnguyen:
Hi @marxjohnson 

FYI, the commit "Set utf-8 encoding in signed URL filename header" is causing 403 error, so I create MRs to revert the commit.

tuanngocnguyen:
Hi @marxjohnson ,

I assume you are working on this, please let me know if it is not the case.
Thanks

marxjohnson:
Hi @tuanngocnguyen,
Thanks for addressing that, I have been on holiday so not had time to look at this further.
That commit was just an attempt to tidy things up, if it was causing problems I believe the original issue should still be resolved without it. I'm not sure that the `utf8_encode()` that was originally there is correct (we are starting with a UTF-8 string, so why encode it as UTF-8 again?) but in my testing it seemed to work as expected whether the filename was ISO-8859-1 only or UTF-8. 

[...]

PR #493, opened 2022-06-15
S3 pre-signed URL: error with unicode filename #455

marxjohnson:
Rebased #456

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -486,9 +486,10 @@ class client extends object_client_base {
     /**
      * @param string $contenthash
      * @param array $headers
+     * @param bool $nicefilename
      * @return signed_url
      */
-    private function generate_presigned_url_s3($contenthash, $headers) {
+    private function generate_presigned_url_s3($contenthash, array $headers = [], $nicefilename = true) {
         $contentdisposition = manager::get_header($headers, 'Content-Disposition');
         if ($contentdisposition !== '') {
             $params['ResponseContentDisposition'] = $contentdisposition;
@@ -503,14 +504,12 @@ class client extends object_client_base {
         $params['Bucket'] = $this->bucket;
         $params['Key'] = $this->bucketkeyprefix . $key;
 
-        $contentdisposition = manager::get_header($headers, 'Content-Disposition');
-        if ($contentdisposition !== '') {
-            $params['ResponseContentDisposition'] = $contentdisposition;
-        }
-
-        $contenttype = manager::get_header($headers, 'Content-Type');
-        if ($contenttype !== '') {
-            $params['ResponseContentType'] = $contenttype;
+        if ($nicefilename) {
+            $result = $this->get_nice_filename($headers);
+            if (!empty($result)) {
+                $params['ResponseContentDisposition'] = $result['Content-Disposition'] . '; ' . $result['filename'];
+                $params['ResponseContentType'] = $result['Content-Type'];
+            }
         }
 
         $command = $this->client->getCommand('GetObject', $params);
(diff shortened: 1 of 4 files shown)


marxjohnson:
@brendanheywood One of our Catalyst EU clients is waiting on this fix, would you mind taking a look now its been rebased?

classes/local/store/s3/client.php
```diff
@@ -486,9 +486,10 @@ public function generate_presigned_url($contenthash, $headers = array()) {
     /**
      * @param string $contenthash
      * @param array $headers
+     * @param bool $nicefilename
      * @return signed_url
      */
-    private function generate_presigned_url_s3($contenthash, $headers) {
+    private function generate_presigned_url_s3($contenthash, array $headers = [], $nicefilename = true) {
```
brendanheywood, line 492: Can you add a unit test for this? It's unclear to see what was wrong with the urls it was creating and how they have changed

marxjohnson:
@brendanheywood I have added unit tests to the branch, and an additional test file for the "Pre-Signed URL Testing" test page. I rebased the branch to put the tests before Phuc's commit, so it's easy to run the tests with and without the change, and see the error when you try and download the file with the unicode character.

tests/object_file_system_test.php
```diff
@@ -585,6 +585,39 @@ public function test_can_generate_signed_url_by_hash_if_object_is_external() {
...
+        }
+    }
+
+    public function test_can_generate_signed_url_with_unicode_filename() {
+        $this->filesystem = new test_file_system();
+        $file = $this->create_remote_file();
+        $filehash = $file->get_contenthash();
+        try {
+            $headers = [
+                    'Content-Disposition' => 'attachment; filename="😀.txt"',
```
scara, line 610: Hello Everyone,
just lurking here 👀.
Is there any reason not to adopt even the [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987#section-3.2) and [RFC 6266](https://www.rfc-editor.org/rfc/rfc6266#section-4.1) indications i.e.:
```
Content-Disposition: attachment; filename*=utf-8''%e2%82%ac%20rates
```
where the _filename_ there is something like `rawurlencode($originalfilename)`?
I mean something like:
```php
$result['filename'] = 'filename="' . utf8_encode($originalfilename) . '"'
                  . "; filename*=utf-8''" . rawurlencode($originalfilename);
```
Maybe it's not required by your use cases right now but it could be future proof.

HTH,
Matteo

marxjohnson: Hi Matteo, you're right, that probably would make sense. With Moodle we can safely assume that the filename is UTF-8 encoded, so specifying that in the header would be correct. 

We are currently doing the rawurlencode() seperately, which doesn't quite follow the pattern you indicated. I will do an additional commit to rework it as you suggested.

scara: > We should be OK just setting the filename as UTF-8 and doing `rawurlencode()`.

Modern browsers should be fine with that, though e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition is not clear enough on the back-compat.
Old benchmarks like http://test.greenbytes.de/tech/tc2231/ give some hints but I think its fine to tell the browsers are now running fine in the modern days 😉.

HTH,
Matteo

tuanngocnguyen:
Reopen this PR as the following commit cause 403 error when retrieve files with pre-signed URL

[Set utf-8 encoding in signed URL filename header](https://github.com/catalyst/moodle-tool_objectfs/pull/493/commits/8e107e6f8eac3a92214d2c056493ae6d66ee91b1)

PR #498, opened 2022-06-29
Update CLOUDFRONT.md - per wr384955

JoshAudette:
Per request in wr384955, these changes resolved issues:
1) The presignedurl_tests.php was giving only content hashes and not original file names -- resolved by changing the caching settings to the Legacy settings; and
2) The presignedurl_tests.php iFrames were giving an error - resolved by disabling the X-Frame-Options in the response headers policy

### DIFF
--- a/CLOUDFRONT.md
+++ b/CLOUDFRONT.md
@@ -97,7 +97,7 @@ cat public_key.pem
 4. _Configure CORS_: disabled
 5. _Strict-Transport-Security_: Enabled, origin override enabled
 6. _X-Content-Type-Options_: Enabled, origin override enabled
-7. _X-Frame-Options_: Enabled, SAMEORIGIN, origin override enabled
+7. _X-Frame-Options_: Disabled
 8. _X-XSS-Protection_: Enabled, block, origin override enabled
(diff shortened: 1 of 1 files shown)


CLOUDFRONT.md
```diff
@@ -112,15 +112,17 @@ cat public_key.pem
...
-10. _Cache key and origin requests_: Choose _Cache policy and origin request policy (recommended)_.
-11. _Cache policy_: Choose CachingOptimized
-12. _Origin request policy_: Choose CORS-S3Origin
-13. _Response headers policy_: Choose CORS-with-preflight-and-SecurityHeadersPolicy-ReadOnly
-14. Click _Create distribution_.
-15. Navigate to https://console.aws.amazon.com/cloudfront/v3/home#/distributions.
-16. Confirm, that _Status_ is _Enabled_ and _Last modified_ is changed from _Deploying_ to the date the distribution was created.
-17. Open your distribution.
-18. Write down _Distribution domain name_ (with https://).
+10. _Cache key and origin requests_: Choose _Legacy cache settings_.
```
brendanheywood, line 115: Some of these worry me that we've missed steps somewhere else and so changing these settings is a workaround to getting it work again. I'm not across the detail so can't know for sure but I think this should be reviewed by others in the au inf team

PR #500, opened 2022-07-01
Revert "Set utf-8 encoding in signed URL filename header"

tuanngocnguyen:
This reverts commit 8e107e6f8eac3a92214d2c056493ae6d66ee91b1.

### DIFF
--- a/classes/local/store/s3/client.php
+++ b/classes/local/store/s3/client.php
@@ -536,7 +536,7 @@ class client extends object_client_base {
             $result = $this->get_nice_filename($headers);
             if (!empty($result)) {
                 $key .= '?response-content-disposition=' .
-                    rawurlencode($result['Content-Disposition']) . ';' . $result['filename'] .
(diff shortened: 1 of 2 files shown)


classes/local/store/s3/client.php
```diff
@@ -604,7 +604,7 @@ private function get_nice_filename($headers) {
 
             if (!empty($originalfilename)) {
                 $result['Content-Disposition'] = $contentdisposition;
-                $result['filename'] = "filename*=utf-8''" . rawurlencode($originalfilename);
+                $result['filename'] = 'filename="' . utf8_encode($originalfilename) . '"';
```
brendanheywood, line 607: no version micro bump?

tuanngocnguyen: Hi @brendanheywood 

I have bump the version, same version as 33 Stable.
This branch will be replaced by 33 Stable so I guess it is not the issue

PR #501, opened 2022-07-01
Revert "Set utf-8 encoding in signed URL filename header"

tuanngocnguyen:
This reverts commit b50c3a0234b3774c4ec8a6d742debc2139b400f5.

### DIFF
--- a/version.php
+++ b/version.php
@@ -25,8 +25,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2021122302;      // The current plugin version (Date: YYYYMMDDXX).
-$plugin->release   = 2021122302;      // Same as version.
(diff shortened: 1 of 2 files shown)


version.php
```diff
@@ -25,8 +25,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2021122302;      // The current plugin version (Date: YYYYMMDDXX).
-$plugin->release   = 2021122302;      // Same as version.
+$plugin->version   = 2021122303;      // The current plugin version (Date: YYYYMMDDXX).
```
brendanheywood, line 28: I am concerned that these two pr's have the same version number which should never happen:

https://github.com/catalyst/moodle-tool_objectfs/pull/502/files#diff-0dc418129946563e59bb69680a5ef0ded4329ed9e56e3d295c25e12ea726726bR28

PR #503, opened 2022-07-13
Switch unicode filename test to 3-byte character

marxjohnson:
MySQL 5.7 uses utf8mb3 encoding by default, meaning 4-byte characters
(such as emoji) are not supported. This ensures compatibility of the
presigned URL testing across databases, and prevents errors when
attempting to visit the settings page.


Peterburnett:
From the MySQL docs:

`You should also be aware that the utf8mb3 character set is deprecated and you should expect it to be removed in a future MySQL release. Please use utf8mb4 instead.`

This feels like an antifeature and instead the database encoding should be updated. The whole point of the page is to test non-standard chars (such as an emoji), so my gut feel is this is somewhat like commenting out a failing test.

brendanheywood:
Following the docs this should be utf8mb4_unicode_ci

https://docs.moodle.org/400/en/MySQL_full_unicode_support

The question I have is what is our github CI doing here and are we testing on this correctly across the board?

brendanheywood:
For reference:

https://github.com/moodlehq/moodle-plugin-ci/blob/d87e9d96dd74f40cb2d93ade8e8d7934775d8ee7/gha.dist.yml#L18-L24

```yaml
      mariadb:
        image: mariadb:10
        env:
          MYSQL_USER: 'root'
          MYSQL_ALLOW_EMPTY_PASSWORD: "true"
          MYSQL_CHARACTER_SET_SERVER: "utf8mb4"
          MYSQL_COLLATION_SERVER: "utf8mb4_unicode_ci"       
```

brendanheywood:
I've set this here
https://github.com/catalyst/catalyst-moodle-workflows/pull/77/files

marxjohnson:
Thanks @Peterburnett ,
Unfortunately this goes beyond the unit test and `presignedurl_tests.php`, the fixture files are used to test the range requests when the plugin settings page is loaded. If the database can't handle the character, that settings page won't load. I agree that the proper long-term fix is to migrate our clients' databases to mb4 collation (or postgres) but that's not an option in the short term, so we need a fix like this to resolve the immediate regression.

The requirement of the fix on this branch wasn't to support emoji specifically, but to support characters outside of the ISO-8859-1 character set, which this still proves.

brendanheywood:
@marxjohnson is this still needed?

marxjohnson:
@brendanheywood Per https://dev.mysql.com/doc/refman/8.0/en/charset.html the default encoding is now mb4, so I think we can close this.

bwalkerl:
Closing as per the most recent comment, which is the required mysql version for for Moodle 4.2+

Issue #505, opened 2022-08-15
Performance - Add filesize to objects table to reduce joins.


andrewmadden:
There are a few queries that would benefit by not needing to join the files table, due to the size the files table can be. If the filesize is added as a field to the objects field, some queries may be able to be simplified. 

This may also solve issue #108 .

Examples:

https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/object_manipulator/candidates/deleter_candidates.php#L37
https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/object_manipulator/candidates/puller_candidates.php#L37
https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/object_manipulator/candidates/recoverer_candidates.php#L37

PR #506, opened 2022-08-15
WIP: issue #505: Add filesize to objects table

andrewmadden:
* Add filesize to objects table
* Create adhoc task that can populate filesize automatically
* Call adhoc in upgrade
* Update SQL queries that no longer need to join files table

https://github.com/catalyst/moodle-tool_objectfs/issues/505

### DIFF
--- a/classes/local/manager.php
+++ b/classes/local/manager.php
@@ -146,10 +146,17 @@ class manager {
                 return $oldobject;
             }
 
+            // Make sure filesize is populated.
+            if (empty($oldobject->filesize)) {
+                $newobject->filesize = $DB->get_field('files', 'filesize', ['contenthash' => $contenthash], IGNORE_MULTIPLE);
+            } else {
+                $newobject->filesize = $oldobject->filesize;
+            }
             return self::update_object($newobject, $newlocation);
         }
         $newobject->timeduplicated = time();
         $newobject->location = $newlocation;
+        $newobject->filesize = $DB->get_field('files', 'filesize', ['contenthash' => $contenthash], IGNORE_MULTIPLE);
         $DB->insert_record('tool_objectfs_objects', $newobject);
 
         return $newobject;
--- a/classes/local/object_manipulator/candidates/deleter_candidates.php
+++ b/classes/local/object_manipulator/candidates/deleter_candidates.php
@@ -34,17 +34,12 @@ class deleter_candidates extends manipulator_candidates_base {
      * @return string
      */
     public function get_candidates_sql() {
-        return 'SELECT MAX(f.id),
-                       f.contenthash,
-                       MAX(f.filesize) AS filesize
-                  FROM {files} f
-                  JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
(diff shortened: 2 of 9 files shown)


classes/local/object_manipulator/candidates/deleter_candidates.php
```diff
@@ -34,17 +34,15 @@ class deleter_candidates extends manipulator_candidates_base {
...
-                       MAX(f.filesize) AS filesize
-                  FROM {files} f
-                  JOIN {tool_objectfs_objects} o ON f.contenthash = o.contenthash
-                 WHERE o.timeduplicated <= :consistancythreshold
-                   AND o.location = :location
-                   AND f.filesize > :sizethreshold
-              GROUP BY f.contenthash,
-                       f.filesize,
-                       o.location';
+        return 'SELECT contenthash,
```
brendanheywood, line 37: The first column must be guaranteed to be unique but we still have a GROUP BY here, can we eliminate the GROUP BY completely now tha we don't have the join?

I think this applies to all the the sql in the same way

classes/task/populate_objects_filesize.php
```diff
@@ -0,0 +1,52 @@
...
+ * @author     Andrew Madden <andrewmadden@catalyst-au.net>
+ * @copyright  2022 Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class populate_objects_filesize extends adhoc_task {
+
+    /**
+     * Action of task.
+     */
+    public function execute() {
```
brendanheywood, line 38: This one worries me as I'd expect it to take forever on a big site.

There is two ways it could be better:
1) rewrite it as pure sql so we don't need to pull data down and push it back

The problem with this is that it's not cross db compliant, BUT we can write two code paths so it does the more performant sql in places where we know it works

```
INSERT INTO {foo} (clientid, domain, usermodified, timecreated, timemodified)
                SELECT id as clientid, domain, usermodified, timecreated, timemodified
                  FROM {bar} client
                 WHERE NOT EXISTS (
                     SELECT 1
                       FROM {foo}
                      WHERE domain = client.domain)"
```

there is precedent for this in core:
https://github.com/moodle/moodle/blob/master/mod/quiz/locallib.php#L1148-L1179

2) The other way is we just page the data and limit it to N updates and then spawn another adhoc task if there are more to do

classes/local/manager.php
```diff
@@ -135,6 +135,7 @@ public static function update_object_by_hash($contenthash, $newlocation) {
         global $DB;
         $newobject = new stdClass();
         $newobject->contenthash = $contenthash;
+        $newobject->filesize = $DB->get_field('files', 'filesize', ['contenthash' => $contenthash], IGNORE_MULTIPLE);
```
brendanheywood: Do we strictly need this here? In theory it should already have a size. The adhoc task should be back filling this data. If we want to deal with the cases where we are touching this table before the adhoc task has completed then lets make it only do this if the filesize is missing.

brendanheywood: Also in some cases the calling code already has the filesize and could optionally pass it in save another db read

PR #512, opened 2022-09-12
Issue505 moodle 33 stable add filesize to objects

andrewmadden:


classes/local/manager.php
```diff
@@ -126,10 +126,11 @@ public static function get_client($config) {
     /**
      * @param $contenthash
      * @param $newlocation
+     * @param int $filesize Size of the file in bytes. Falls back to stored value if not provided.
      * @return mixed|stdClass
      * @throws \dml_exception
      */
-    public static function update_object_by_hash($contenthash, $newlocation) {
+    public static function update_object_by_hash($contenthash, $newlocation, $filesize = 0) {
```
brendanheywood: I assume we need to be careful around filesize = null meaning we don't yet know it vs filesize = 0 meaning we have a file and its size is 0?

classes/local/manager.php
```diff
@@ -139,22 +140,28 @@ public static function update_object_by_hash($contenthash, $newlocation) {
             $newobject->timeduplicated = $oldobject->timeduplicated;
             $newobject->id = $oldobject->id;
 
-            // If location hasn't changed we do not need to update.
-            if ((int)$oldobject->location === $newlocation) {
+            // If location hasn't changed we do not need to update unless filesize is not populated.
+            if ((int)$oldobject->location === $newlocation && !empty($oldobject->filesize)) {
```
brendanheywood: similar to above I think we should not use empty we should use === null everywhere

classes/local/manager.php
andrewmadden: To make the changes as safe as possible, I set the DB field to default to 0 so the filesize will never be null. I mainly did this as some queries use int comparators (filesize > 0) and I thought null may break these queries.

https://github.com/catalyst/moodle-tool_objectfs/pull/512/files#diff-4013433adb48858dfeac13caf181742020b6c693bc853298312c45b134377cb6R13

So I could set the default for this param to be null, but it won't be null for the old object.

andrewmadden: Just doing some quick googling. Looks like I may have been overcautious and NULL SQL values will just be ignored when using < > <= >= operators. I think it's worth the refactor so we don't keep checking files that have a filesize of 0

PR #515, opened 2022-09-26
Issue505 moodle 310 stable add filesize to objects

andrewmadden:
* Add filesize to objects table
* Create adhoc task that can populate filesize automatically
* Call adhoc in upgrade
* Update SQL queries that no longer need to join files table
* Initiate filesize when creating an object


keevan (approved):
Code changes are identical to https://github.com/catalyst/moodle-tool_objectfs/pull/512/files

So it seems good to merge into this branch also.

PR #518, opened 2022-11-07
issue505 Do not include objects in population task lookup that are orphaned or errors

### LINKED ISSUE #505: Performance - Add filesize to objects table to reduce joins.

andrewmadden:
Closes #505

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -154,5 +154,15 @@ function xmldb_tool_objectfs_upgrade($oldversion) {
 
         upgrade_plugin_savepoint(true, 2021122306, 'tool', 'objectfs');
     }
+
+    if ($oldversion < 2021122307) {
+        // Check to make sure adhoc task not already running.
+        if (!$DB->record_exists('task_adhoc', ['classname' => '\tool_objectfs\task\populate_objects_filesize'])) {
+            // Populate the filesize field.
+            \core\task\manager::queue_adhoc_task(new \tool_objectfs\task\populate_objects_filesize());
+        }
+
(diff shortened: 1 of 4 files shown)


brendanheywood:
I think this is good but lets hold off on the merge till we've battle tested it at scale

db/upgrade.php
```diff
@@ -154,5 +154,15 @@ function xmldb_tool_objectfs_upgrade($oldversion) {
 
         upgrade_plugin_savepoint(true, 2021122306, 'tool', 'objectfs');
     }
+
+    if ($oldversion < 2021122307) {
+        // Check to make sure adhoc task not already running.
+        if (!$DB->record_exists('task_scheduled', ['classname' => '\tool_objectfs\task\populate_objects_filesize'])) {
```
brendanheywood: this is the wrong table, should be task_adhoc

andrewmadden: good catch

PR #520, opened 2022-11-09
issue505 Do not include objects in population task lookup that are orphaned or errors

andrewmadden:


danmarsden:
@andrewmadden is this waiting in the queue for any good reason? - I've just had a site showing this task running really slow and this seems to drop the number of records it's dealing with by 25%  so I'd be keen to merge it unless you have any objections?

classes/task/populate_objects_filesize.php
```diff
@@ -43,10 +43,12 @@ public function execute() {
         $maxupdates = !empty($data->maxupdates) ? $data->maxupdates : self::MAX_UPDATES;
 
         // Get all objects without a filesize and join them to a filesize from the files table.
+        // Values less than 0 for object's location indicate an error for the object.
         $sql = "SELECT o.id, o.contenthash, o.timeduplicated, o.location, f.filesize
                   FROM {tool_objectfs_objects} o
              LEFT JOIN {files} f ON o.contenthash = f.contenthash
```
danmarsden, line 49: shouldn't this be a "Join" instead of a Left Join? - if the point is to get the filesize from the files table, it shouldn't return records when there isn't a files record for the item?

danmarsden:
merged this in and dropped a change on top to use a join instead of a left join. - When the number of orphaned records is higher than 100,000 (returning a null filesize) - this ad-hoc task never ends, - it runs on the first 100,000 records, then spawns another ad-hoc task and does the same thing all over again....

andrewmadden:
> merged this in and dropped a change on top to use a join instead of a left join. - When the number of orphaned records is higher than 100,000 (returning a null filesize) - this ad-hoc task never ends, - it runs on the first 100,000 records, then spawns another ad-hoc task and does the same thing all over again....

@danmarsden Sorry for delay. This was a forward port of https://github.com/catalyst/moodle-tool_objectfs/pull/518 that was merged a few months ago. Looks like it slipped under the radar after we checked the performance bug was fixed in the 33 branch. Thanks for pushing it along.

The branches will be slightly out of sync with your change to the SQL but I don't think it will be an issue.

PR #522, opened 2022-12-02
Moodle 4.1 testing fixes

### LINKED ISSUE #523: Test support for 4.1 -> Some issues found
I have tested installation, automated tests and signed URL tests on Moodle 4.1. I found the following minor issues:

* The unit test classes need renaming, as PHPUnit has deprecated class name that do not match the file name.
* Test data generator methods are prefixed with `test_` so are run as tests and failing as they make no assertions.
* Some other tests are failing as they make no assertions.
* Unrelated to 4.1 but still worth fixing, some admin settings are missing defaults, causing warnings on the admin screens.
* Some tests assume no files exist by default, so are failing due to incorrect record counts.

marxjohnson:
Fixes #523

This mainly fixes issues with unit tests using deprecated class names or failing due to tests not performing assertions.
There was also an issue with some tests assuming that no files exist on install, but assignfeedback_editpdf creates 5 files on installation.

There were also some warnings showing up on the admin screens due to missing default values for plugin settings. Not strictly 4.1-related, but worth fixing.


danmarsden:
thanks @marxjohnson mostly +1 from me - I'm just on the fence about the set_config call in the install.php file though - I'd quite like to delete that install.php file as it contains stuff for Moodle 3.6 and lower and the current branch is for MOODLE 3.10 and higher.

can we add improve the location where expirationtime is used to better check if it is set instead?

danmarsden:
I've just cherry-picked the main components of this into #528 excluding that change for the missing default values - @marxjohnson if you have a moment feel free to add a new commit to fix those up that doesn't need to use an install.php file! - thanks!

PR #525, opened 2022-12-06
Fix #524 - Don't hit external storage during normal file upload

### LINKED ISSUE #524: Performance: Uploading a file to Moodle shouldn't hit object storage during the upload process
Basically just summarising the issue from our internal WR#393527 to allow for a public reason for the change I'm proposing...

Turning object storage on has a significant impact on file upload performance for us when dealing with SCORM packages - we don't use AWS but have an openstack instance which doesn't quickly return a 404 response when asking about a file that doesn't exist in the external storage.

When you upload a new SCORM package it extracts the files then calls add_file_from_string for each file in the zip. the Object storage class overrides this function so it can ask the external storage "do you have the file yet" and if not, adds a record to the tool_objectfs_objects to make sure it's dealt with.
https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/object_file_system.php#L882

The problem is that the "do you have a file yet" when the file doesn't exist in the external storage device (it's a new SCORM package) can be really slow on our end and what would usually take 1min to upload (when just touching local disk) blows out to something crazy like 2min 30secs just because of all the "do you have the file yet" calls against the external storage.
Some scorm packages have thousands of small files - and in many cases those small files will be under the sizethreshold set by the objectfs plugin so will never end up in external storage anyway.

The reality is that this check isn't needed at all, because the objectfs plugin has a cron task that looks for new files that don't exist in the external storage and uploads them 

Some local testing with our infrastructure showed the following results when uploading a SCORM package with lots of files:
Without object storage and local disk: 0:09 minutes
With object storage and local disk: 6:24 minutes

We could add a setting to turn this on/off, but I'm curious to hear from anyone testing with a large SCORM package on AWS as I expect that the change I'm proposing might improve performance there too.

(PR incoming)

danmarsden:
See issue #524 for more background, but this prevents the plugin from hitting the external storage system during a normal file upload - and just relies on the scheduled task to manage the process completely instead.

The alternative here is to add a setting to turn this behaviour on/off but I'm not sure why we need to do this during the user file upload at all and we should just be able to rely on the scheduled tasks.

Passing to @brendanheywood to review - but happy for any other feedback from other interested parties too! :-)

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -855,37 +855,6 @@ abstract class object_file_system extends \file_system_filedir {
     }
 
     /**
-     * Add the supplied file to the file system and update its location.
-     *
-     * @param string $pathname Path to file currently on disk
-     * @param string $contenthash SHA1 hash of content if known (performance only)
-     * @return array (contenthash, filesize, newfile)
-     */
-    public function add_file_from_path($pathname, $contenthash = null) {
-        $result = parent::add_file_from_path($pathname, $contenthash);
-
-        $location = $this->get_object_location_from_hash($result[0]);
-        manager::update_object_by_hash($result[0], $location, $result[1]);
-
-        return $result;
-    }
-
-    /**
-     * Add a file with the supplied content to the file system and update its location.
-     *
-     * @param string $content file content - binary string
-     * @return array (contenthash, filesize, newfile)
-     */
-    public function add_file_from_string($content) {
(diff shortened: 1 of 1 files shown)


brendanheywood:
+1 from me, the main thought I had was around trying to update the metadata location without doing the read, but it's fine to just leave it all the cron to clean up

PR #526, opened 2022-12-06
tool_objectfs: optimise token usage for swift

danmarsden:
Cache swift token so it's not generated on every request.


danmarsden:
Matt wrote this, I did the PR, and it's been running in prod for a few weeks on a couple of our sites - seems to be going well so I'm pushing this up now.

PR #536, opened 2023-01-16
Issue519 moodle 310 stable presigned urls check

andrewmadden:


andrewmadden:
Two examples of notifications in the settings:

Screenshot from 2023-01-31 01-14-31
Screenshot from 2023-01-31 01-13-42

Can't give example of error notification without deliberately breaking the code.

classes/check/presigned_urls.php
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * @return action_link
+     */
+    public function get_action_link(): ?\action_link {
+        return new \action_link(
+            new \moodle_url('/admin/settings.php?section=tool_objectfs_settings'),
+            get_string('pluginsettings', 'tool_objectfs'));
+    }
+
+    /**
```
brendanheywood, line 47: Add a details page which just dumps out the list of plugin file urls for each fixture so you can manually retry each one

PR #537, opened 2023-01-25
Fixing lock release error

caiovncius:
In Moodle 3.11.+ we getting this error: 

`A lock was created but not released at: [dirroot]/admin/tool/objectfs/classes/local/store/object_file_system.php on line 216 Code should look like: $factory = \core\lock\lock_config::get_lock_factory('type'); $lock = $factory->get_lock(50e571f912ea1e8d29ccfeb56a43748b59f24c3d); $lock->release(); // Locks must ALWAYS be released like this.`

In my scenario, looks it´s happening when I try access objectfs in interface and cron in same time.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -214,6 +214,7 @@ abstract class object_file_system extends \file_system_filedir {
         $lockfactory = \core\lock\lock_config::get_lock_factory('tool_objectfs_object');
         $this->logger->start_timing();
         $lock = $lockfactory->get_lock($resource, $timeout);
+        $lock->release();
         $this->logger->end_timing();
         $this->logger->log_lock_timing($lock);
 


classes/local/store/object_file_system.php
```diff
@@ -214,6 +214,7 @@ public function acquire_object_lock($contenthash, $timeout = 0) {
         $lockfactory = \core\lock\lock_config::get_lock_factory('tool_objectfs_object');
         $this->logger->start_timing();
         $lock = $lockfactory->get_lock($resource, $timeout);
+        $lock->release();
```
dvdcastro, line 217: This is not correct. You can't release the lock as soon as you get it. There must be a reason why the lock is not being released. We also found this bug ourselves.

dvdcastro (changes requested):

brendanheywood:
Closing this as this is the wrong approach

PR #542, opened 2023-02-21
Corrections to the cloudfront documentation

petermAprende:
I see 2 issues with the cloudfront documentation:

### _X-Frame-Options_: Same Origin

This breaks any assets loading in via iframe (such as for scorms)

### _Origin request policy_: CORS-S3Origin

The stock CORS-S3Origin policy doesn't forward the content disposition content type headers. Without these headers, files will be forced to download with their hash as their name.


petermAprende:
Fixes #546

twhateley (approved):
I have tested these changes and they were necessary and worked for me.

Issue #543, opened 2023-02-28
The pre-signed URL check triggers on every admin page


petermAprende:
At the moment, the pre-signed URL check (which fires off several curl requests) is triggered anytime `$ADMIN->fulltree && $signingsupport`. This means, triggering several curl requests loading any other admin pages.

petermAprende:
Just that this was only happening because I was on the branch created for #519

PR #545, opened 2023-02-28
fix: only trigger pre-signed check if on tool_objectfs settings page

petermAprende:
Fixes catalyst/moodle-tool_objectfs#543


brendanheywood:
I am going to close this, we have a better alternative in planning which replaces this with an async check API based check

Issue #547, opened 2023-03-01
Stop using the legacy cron function


michael-milette:
This request is to integrate the change for the issue described in:
https://github.com/catalyst/moodle-tool_objectfs/issues/418

Notice: Use of legacy cron is deprecated (tool_objectfs_cron). Please use scheduled tasks.* line 2944 of /lib/deprecatedlib.php: call to debugging()
* line 172 of /lib/classes/task/legacy_plugin_cron_task.php: call to cron_execute_plugin_type()
* line 253 of /lib/cronlib.php: call to core\task\legacy_plugin_cron_task->execute()
* line 120 of /lib/cronlib.php: call to cron_run_inner_scheduled_task()
* line 73 of /lib/cronlib.php: call to cron_run_scheduled_tasks()
* line 81 of /admin/cron.php: call to cron_run()
 in /var/www/moodle/lib/weblib.php on line 3268

The Notice appears when cron triggers the Moodle scheduled tasks.

A PR was previously submitted to address this issue. However, it was never integrated into the MOODLE_310_STABLE branch.:
https://github.com/catalyst/moodle-tool_objectfs/commit/dd33bc8eef5fb005793bbd952d173a4ce7da0132

Best regards,

Michael Milette

michael-milette:
Note: #548 updates the fix #418 to be compatible with the MOODLE_310_STABLE branch of tool_objectfs.

PR #548, opened 2023-03-02
Fix-547: Eliminated legacy cron error.

michael-milette:
This is to address the issue described in:
https://github.com/catalyst/moodle-tool_objectfs/issues/547


brendanheywood:
This was already fixed elsewhere

PR #555, opened 2023-03-22
Set metadata location on upload

### LINKED ISSUE #554: Uploading a new file should still set its location metadata to *something*, even if not perfect
This is a regression caused by #524  (fyi @danmarsden )

Previous state:

We used to upload a file, and then see if that file existed in the remote store, and then based on that set its location to either remote or local. 

Current state:

We do nothing at all which means the file is uploaded without any objectfs metadata. This means we rely on the check_objects_location scheduled task to back fill the data. check_objects_location is an extremely expensive task to run on a large site, we basically only really want to run that to catch very weird edge cases that aren't getting caught as they are uploaded, which is no longer happening. On our sites we have throttled this task right back to run once a week or even once a month. Now we get the unintended side effect that because this task hasn't run, objectfs doesn't know about the files which means they site in site data building up and then once a month get shifted and this has only started happening since this code has progressively rolled out.

Proposed state:
When we upload a file, we still don't want to touch the object store which means we can't correctly determine if the file is actually there or not. But we can do some heuristics and make a pretty good guess:

1) worst case we can do no heuristic and just say the file is local. If the file is also remote then it doesn't matter and the metadata will get corrected when it tries to sync it and it is already there
2) we can look up for the existence of any other files which have the same hash and copy its metadata. This should be correct but on the off chance it is wrong then it was already wrong, and the various scheduled tasks will eventually fix it up anyway

brendanheywood:
Closes #554

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -995,4 +995,40 @@ abstract class object_file_system extends \file_system_filedir {
         debugging('Objectfs does not implement a trashdir.');
         return '';
     }
+
+    /**
+     * Add the supplied file to the file system and update its location.
+     *
+     * @param string $pathname Path to file currently on disk
+     * @param string $contenthash SHA1 hash of content if known (performance only)
+     * @return array (contenthash, filesize, newfile)
+     */
+    public function add_file_from_path($pathname, $contenthash = null) {
+        $result = parent::add_file_from_path($pathname, $contenthash);
+
+        // Rather than getting its exact location we just set it to local.
+        // Almost all file uploads will be unique, and if it is a duplicate
+        // then this will be corrected when the file is synced later.
+        manager::update_object_by_hash($result[0], OBJECT_LOCATION_LOCAL, $result[1]);
+
+        return $result;
+    }
+
+    /**
+     * Add a file with the supplied content to the file system and update its location.
+     *
+     * @param string $content file content - binary string
+     * @return array (contenthash, filesize, newfile)
+     */
+    public function add_file_from_string($content) {
+        $result = parent::add_file_from_string($content);
+
+        // Rather than getting its exact location we just set it to local.
+        // Almost all file uploads will be unique, and if it is a duplicate
+        // then this will be corrected when the file is synced later.
+        manager::update_object_by_hash($result[0], OBJECT_LOCATION_LOCAL, $result[1]);
+
+        return $result;
+    }
+
 }
--- a/version.php
+++ b/version.php
@@ -25,8 +25,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2023013101;      // The current plugin version (Date: YYYYMMDDXX).
-$plugin->release   = 2023013100;      // Same as version.
+$plugin->version   = 2023051700;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->release   = 2023051700;      // Same as version.
 $plugin->requires  = 2020110900;      // Requires Filesystem API.
 $plugin->component = "tool_objectfs";
 $plugin->maturity  = MATURITY_STABLE;


classes/local/store/object_file_system.php
```diff
@@ -995,4 +995,40 @@ protected function get_trash_fullpath_from_hash($contenthash) {
         debugging('Objectfs does not implement a trashdir.');
         return '';
     }
+
+    /**
+     * Add the supplied file to the file system and update its location.
```
keevan, line 1000:
```suggestion
     * Add the supplied file to the file system and update its object location.
```
To me, location is too generic of a word. I've provided a suggestion but happy for any alternatives that might make it clearer.

classes/local/store/object_file_system.php
```diff
@@ -995,4 +995,40 @@ protected function get_trash_fullpath_from_hash($contenthash) {
...
+     * @param string $contenthash SHA1 hash of content if known (performance only)
+     * @return array (contenthash, filesize, newfile)
+     */
+    public function add_file_from_path($pathname, $contenthash = null) {
+        $result = parent::add_file_from_path($pathname, $contenthash);
+
+        // Rather than getting its exact location we just set it to local.
+        // Almost all file uploads will be unique, and if it is a duplicate
+        // then this will be corrected when the file is synced later.
+        manager::update_object_by_hash($result[0], OBJECT_LOCATION_LOCAL, $result[1]);
```
keevan, line 1012: Could we do something like this:
```suggestion
        [$contenthash, $filesize] = parent::add_file_from_path($pathname, $contenthash);

        // Rather than getting its exact location we just set it to local.
        // Almost all file uploads will be unique, and if it is a duplicate
        // then this will be corrected when the file is synced later.
        manager::update_object_by_hash($contenthash, OBJECT_LOCATION_LOCAL, $filesize);
```

Based on the return array of add_file_from_path. Similar comment for the other function too.
Note the concern would be the duplicate contenthash variable name, where dual purpose usage is generally discouraged, looking at the implementation, it seems to update and return this value anyways.
Ref: https://github.com/moodle/moodle/blob/9ee4f8db8ba7f9c3596592a9e14c05c1ea5a024b/lib/filestorage/file_system_filedir.php

keevan: Actually I saw the return value for this function, and I'm retracting my initial suggestion.

classes/local/store/object_file_system.php
```diff
@@ -995,4 +995,40 @@ protected function get_trash_fullpath_from_hash($contenthash) {
...
+
+        return $result;
+    }
+
+    /**
+     * Add a file with the supplied content to the file system and update its location.
```
keevan, line 1018:
```suggestion
     * Add a file with the supplied content to the file system and update its object location.
```
Similar to a previous suggested change.

classes/local/store/object_file_system.php
```diff
@@ -995,4 +995,40 @@ protected function get_trash_fullpath_from_hash($contenthash) {
...
+     * @param string $contenthash SHA1 hash of content if known (performance only)
+     * @return array (contenthash, filesize, newfile)
+     */
+    public function add_file_from_path($pathname, $contenthash = null) {
+        $result = parent::add_file_from_path($pathname, $contenthash);
+
+        // Rather than getting its exact location we just set it to local.
+        // Almost all file uploads will be unique, and if it is a duplicate
+        // then this will be corrected when the file is synced later.
+        manager::update_object_by_hash($result[0], OBJECT_LOCATION_LOCAL, $result[1]);
```
keevan, line 1012:
```suggestion
        [$contenthash, $filesize] = $result;
        manager::update_object_by_hash($contenthash, OBJECT_LOCATION_LOCAL, $filesize);
```
With the preference of using clear names (instead of numerical indexes)

classes/local/store/object_file_system.php
```diff
@@ -995,4 +995,40 @@ protected function get_trash_fullpath_from_hash($contenthash) {
...
+     * @param string $content file content - binary string
+     * @return array (contenthash, filesize, newfile)
+     */
+    public function add_file_from_string($content) {
+        $result = parent::add_file_from_string($content);
+
+        // Rather than getting its exact location we just set it to local.
+        // Almost all file uploads will be unique, and if it is a duplicate
+        // then this will be corrected when the file is synced later.
+        manager::update_object_by_hash($result[0], OBJECT_LOCATION_LOCAL, $result[1]);
```
keevan, line 1029:
```suggestion
        [$contenthash, $filesize] = $result;
        manager::update_object_by_hash($contenthash, OBJECT_LOCATION_LOCAL, $filesize);
```

brendanheywood:
ci is broken but its not related to this

Issue #65, opened 2017-05-01
watch MDL-58281


kenneth-hendricks:

kenneth-hendricks:
Changes will most likely have to be made to the objectfs_file_system

brendanheywood:
I've just pushed a new patch for:

https://tracker.moodle.org/browse/MDL-58281

and yes a patch will be needed here to

brendanheywood:
MDL-58281 landed a while back, not sure what is left to do on this really

PR #562, opened 2023-05-04
throw exception if readfile unsuccessful.

x-iy:
References issue https://github.com/catalyst/moodle-tool_objectfs/issues/65

https://wrms.catalyst.net.nz/wr.php?request_id=406372

Throw exception if readfile fails.

Similar to https://github.com/moodle/moodle/blob/master/lib/filestorage/file_system.php#L45


brendanheywood:
Good to go when green (I've also turned on automerge)

keevan (approved):

PR #568, opened 2023-06-19
replaced deprecated function cron_trace_time_and_memory() for Moodle 4.2+

opitz:
As this change is required for Moodle 4.2+ only I suggest to create a new branch (e.g. MOODLE_402_STABLE)


danmarsden:
This is really just removing technical debt - this code is only used in really old Moodle sites and the current MOODLE_310_STABLE branch doesn't provide support for those branches anyway - I don't think we need to create a new branch as this looks safe to merge into the current branch.

aspark21:
A MOODLE_402_STABLE branch has been cut so it might be worth merging either this or #576

bwalkerl:
Closing as this was removed in #576

PR #570, opened 2023-07-10
add during_initial_install check in update_object_by_hash function

ghost:


classes/local/manager.php
```diff
@@ -139,6 +139,10 @@ public static function update_object_by_hash($contenthash, $newlocation, $filesi
         $newobject = new stdClass();
         $newobject->contenthash = $contenthash;
 
+        if (during_initial_install()) {
```
brendanheywood, line 140: Why not do this first above line 139?

PR #576, opened 2023-08-23
Removed legacy cron function

sammarshallou:
In Moodle 4.2, cron_trace_time_and_memory has been replaced by a different function in a namespaced location, and you get deprecation warnings for including cronlib.php as well as for calling the function.

This change removes the old cron function in lib.php, and changes the code that calls cron_trace_time_and_memory so that it uses the old method prior to 4.2, and the new one after.

Note: This is a duplicate of #568 but I wrote the code before I saw that one! While it is safe in the MOODLE_310_STABLE branch, to remove the cron function from lib.php (as 3.10 already supported task API), it is not safe to change the cron_trace_time_and_memory function because the new \core\cron class does not exist in 3.10, or any version until 4.2. So I added a version check.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -69,22 +69,6 @@ define('TOOL_OBJECTFS_DELETE_EXTERNAL_NO', 0);
 define('TOOL_OBJECTFS_DELETE_EXTERNAL_TRASH', 1);
 define('TOOL_OBJECTFS_DELETE_EXTERNAL_FULL', 2);
 
-// Legacy cron function.
-function tool_objectfs_cron() {
-    mtrace('RUNNING legacy cron objectfs');
-    global $CFG;
-    if ($CFG->branch <= 26) {
-        // Unlike the task system, we do not get fine grained control over
-        // when tasks/manipulators run. Every cron we just run all the manipulators.
-        (new manipulator_builder())->execute_all();
-
-        \tool_objectfs\local\report\objectfs_report::cleanup_reports();
-        \tool_objectfs\local\report\objectfs_report::generate_status_report();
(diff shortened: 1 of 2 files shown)


brendanheywood:
thanks @sammarshallou - I think this is good to land without a new stable. Not sure about the red ci that seems unrelated

brendanheywood:
@sammarshallou does the 'Could not use "Hint_ResultPrinter" as printer: class does not exist' error mean anything to you?

sammarshallou:
@brendanheywood I saw it but sorry, no, I don't know how the CI infrastructure here is supposed to work and I don't understand it at all! I presume these errors are unrelated to the change here... 

I did a quick search and found this issue in which remoing that class (or one with the same name...) was discussed because it apparently wasn't needed, but that was done way back in Moodle 3.10 so probably not related https://tracker.moodle.org/browse/MDL-67673

It looks like, at least in my repository, the same error occurs if I just push the MOODLE_310_STABLE branch without my change.

https://github.com/sammarshallou/moodle-tool_objectfs/actions/runs/5961359997

aspark21:
the  change to the phpunit.xml was applied last week so with a rebase those errors should clear up.

sammarshallou:
@aspark21 Thanks! I've rebased, let's see.

sammarshallou:
Somebody already just deleted the trace_time_and_memory call in the MOODLE_402_STABLE branch, so I've changed this PR to remove that part and just delete the unused cron function and corresponding test. (I'm not sure if the existence of this function actually causes problems but it's a bit pointless.)

PR #583, opened 2023-10-26
[#582] fix namespace for exception class

### LINKED ISSUE #582: Exception - Class 'tool_objectfs\local\store\s3\file_exception' not found
https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/s3/file_system.php#L64

Looks like this line should be `throw new \file_exception('storedfilecannotreadfile', $file->get_filename());`

It causes an error currently.

jwalits:
Fixes #582

### DIFF
--- a/classes/local/store/s3/file_system.php
+++ b/classes/local/store/s3/file_system.php
@@ -61,7 +61,7 @@ class file_system extends object_file_system {
 
         if (!$success) {
             manager::update_object_by_hash($file->get_contenthash(), OBJECT_LOCATION_ERROR);
-            throw new file_exception('storedfilecannotreadfile', $file->get_filename());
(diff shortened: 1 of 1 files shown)


x-iy:
Would this also be a namespace issue?
https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/object_file_system.php#L236

jwalits:
> Would this also be a namespace issue? https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/object_file_system.php#L236

Yeh, that also looks like it should be `\file_exception(..` though not sure, how to force an error on that code

x-iy:
Looks good to me

Issue #579, opened 2023-08-31
Test range request failed - Minio and Moodle S3 Storage connection


Frostist:
<img width="917" alt="image" src="https://github.com/catalyst/moodle-tool_objectfs/assets/37545822/9e366ce3-c227-4889-a101-0cda01f81673">

Please can anyone provide some help with this.

I have connected a Minio Bucket server to Moodle and the Minions Console has reported that the test data has been uploaded but Moodle is complaining...

If I go to this link in the photo above, I get taken to the blender bunny video.

Frostist:
<img width="703" alt="image" src="https://github.com/catalyst/moodle-tool_objectfs/assets/37545822/1311fcc6-b660-43d8-a1b6-99294170a4af">

This is the bucket server on Minio

Frostist:
Note to others that have found this to be an issue!

go to:
Site Administration -> General -> Http Security

Then remove 127.0.0.1 and localhost off the cURL blocked hosts list
and 
add the port 9000 to the cURL allowed ports list.

Done!

Frostist:
For others that are trying to get Minio and Moodle Objectfs S3 storage to work together.

Here are the details that I used.

<img width="1251" alt="image" src="https://github.com/catalyst/moodle-tool_objectfs/assets/37545822/97da9ba7-6776-4278-abed-1442ec2e0aa3">

Issue #374, opened 2020-08-18
Minio.io Client Integration


andreigae:
Hi i Would like to know it you have any plan to integrate minio.io server with this plugin (minio.io = a open source s3 object server )

Thanks

danmarsden:
From what I understand it uses the same API as Amazon S3 - so it might even "just work" ?

otherwise - no active plans that I'm aware of specifically targeting minio - pull requests are always welcome, otherwise If you would like to sponsor work on this please contact us here: https://www.catalyst-au.net/contact-us

brendanheywood:
We support Google GCS which is compatible with S3 with an alternate url, so yes in theory this should just work.

https://github.com/catalyst/moodle-tool_objectfs#google-gcs

choma:
I wasn't able to get it working with minio. I'm not sure, but I guess it has to do with the fact that minio does not provide ACL (just like DigitalOcean Spaces), so maybe it would work if we had an option to edit the base URL on DigitalOcean implementation.

If anybody get it working, please share your settings. It would be a great option to test things locally, avoiding having to pay a service.

rendihasan:
hello, @andreigae have you successfully integrated with minio?

EsdrasCaleb:
So  I configured the minIO here I followerd the same configuration of the google
https://github.com/catalyst/moodle-tool_objectfs#google-gcs
Using the key and secret that I generated and the bucket that I generated.
The only caverat is that I need to use an IP:8000 in the s3_base_url


I cant read files. It starts with s3

If anyone could help me in config...

EsdrasCaleb:
I cant read files it starts with s3

Frostist:
@rendihasan I managed to get it working.

https://github.com/catalyst/moodle-tool_objectfs/issues/579

Frostist:
> So I configured the minIO here I followerd the same configuration of the google https://github.com/catalyst/moodle-tool_objectfs#google-gcs Using the key and secret that I generated and the bucket that I generated. The only caverat is that I need to use an IP:8000 in the s3_base_url
> 
> I cant read files. It starts with s3
> 
> If anyone could help me in config...

Use the AWS Plugin instead.

PR #590, opened 2024-01-08
Update README.md for Minio.io Settings

Frostist:
Added Minio.io Documentation as per my issue report. https://github.com/catalyst/moodle-tool_objectfs/issues/579

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ A remote object storage file system for Moodle. Intended to provide a plug-in th
 * [Currently supported object stores](#currently-supported-object-stores)
   * [Roadmap](#roadmap)
   * [Amazon S3](#amazon-s3)
+  * [Min.io S3](#minio-s3)
   * [Google gcs](#google-gcs)
   * [Azure Blob Storage](#azure-blob-storage)
   * [DigitalOcean Spaces](#digitalocean-spaces)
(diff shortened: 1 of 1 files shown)


Frostist:
As per my conversation found https://github.com/catalyst/moodle-tool_objectfs/issues/579
 and https://github.com/catalyst/moodle-tool_objectfs/issues/374

I wanted to update the community to the fact they can use a 3rd party service like Min.io for a open source S3 bucket service.

danmarsden:
Thanks for the PR!

I think before we approve this we should make it clearer in the readme what is "supported" vs what has been "tested as working"

Catalyst doesn't really "support" digital ocean or Min.io - we'd rely on community support for testing/fixing issues with using those.

@brendanheywood - that make sense to you too?

Frostist:
That's fair enough, I just thought it would be nice for users to see they can setup their own minio.io server and DYI the s3 storage instead of going the AWS or Google route.

Thanks!

danmarsden:
yeah I think it makes sense to add the details to the readme, I just don't like the title in the readme that suggests "we" support it directly - along with some other options that I don't think we really "support" ourselves.

Frostist:
Okay! Is there a different section of the readme where I could maybe put it then?

danmarsden:
nah - I think we probably need to adjust the section a bit on our end to make it clearer what "support" means - in particular we offer our code free to download, but don't actively test against all those in the list. I suspect Brendan is the best person to rewrite that section but if he doesn't jump in here I'll see if I can put something together sometime... :-)

brendanheywood:
Yeah and in hindsite I wish we gone with sub plugins so that all these other file stores can be managed by different maintainers in their own repo.

I think maybe have a 'supported' section vs 'community add ons', the first would be the things we personally use in production which is S3, Openstack and Azure but even that it murky as we may not be deploying those for all moodle stables.

danmarsden:
as a first step, I'm going to merge this in and just use github editing tools to update the readme a bit more to make it a little clearer that Catalyst doesn't fully "support" them directly.

Issue #480, opened 2022-04-27
Add check API


dmitriim:
As this plugin can be a critical path to the inf set up, would be useful to add some health check using check API.

Candidates:

- [ ] proxyrangerequests is anbled, but "Test range request failed"
- [ ] A number of missing files are growing 
- [ ] ....
- [ ] ....

brendanheywood:
+1 but lets not worry too much about getting them all done in this issue, we can split them up into different checks and issues

dmitriim:
Yes. We can throw more ideas here and then tackle them separately.

brendanheywood:
+1 to have a super basic check which is that it can both connect to the store, and do something simple like a head request for a file which is knows is in the store to make sure it works end to end

PR #592, opened 2024-01-12
WIP check api

matthewhilton:
A work in progress update to add some basic checks to objectfs - related to #480 

Basically the idea is to expose the current methods we have available in `object_client` via the check api https://moodledev.io/docs/apis/subsystems/check

1. `test_permissions`
2. `test_connection`
3. `test_range_request`

As well as add a new method:
4. `test_configuration` to detect any config level problems e.g. missing keys, missing config, etc...

1,2,3 are basically the same code, just calling a different method, so these share a check API class. 3. is only enabled if range requests are enabled.

4 is special since it runs only if the `$CFG->alternative_file_system` is set, whereas 1,2,3 only run if the configuration is OK (i.e. 4 is passing)

Note this only targets `MOODLE_310_STABLE` since lower versions (< 3.9) do not have the check API.

**TODO**
- [ ] Improve consistency in return value of `test_*` functions - return check results.
- [ ] Delete check/proxy_range_request ? Seems redundant now.
- [ ] Check for usages of `test_*` methods to try to avoid any regressions with the function signature change.
- [ ] Look at replacing `get_client_checks` and make them all use the check API
- [ ] Check performance - should we run these all the time or only some of the time ?
- [ ] PHPDoc + phpcs + lang strings
- [ ] Unit tests ?


classes/check/store.php
```diff
@@ -0,0 +1,122 @@
...
+
+                case self::TYPE_RANGEREQUEST:
+                    return $client->test_range_request(new $config->filesystem());
+
+                case self::TYPE_PERMISSIONS:
+                    return $client->test_permissions(false);
+            }
+        } catch (Throwable $e) {
+            // Usually the SDKs will throw exceptions if something doesn't work, so we want to catch these.
+            return new result(result::CRITICAL, get_string('check:storecheck:error', 'tool_objectfs')
```
matthewhilton: `CRITICAL` or `ERROR` ?

classes/local/store/azure/client.php
```diff
@@ -202,25 +203,23 @@ protected function get_filepath_from_hash($contenthash) {
...
-            $connection->details = $e->getMessage();
+            return new result(result::ERROR, get_string('check:failed', 'tool_objectfs'), $e->getMessage());
         }
 
-        return $connection;
+        return new result(result::OK, get_string('check:passed', 'tool_objectfs'));
     }
 
-    public function test_permissions($testdelete) {
+    public function test_permissions($testdelete): result {
```
matthewhilton, line 226: needs phpdoc

classes/local/store/object_client_base.php
```diff
@@ -89,25 +91,40 @@ public function generate_presigned_url($contenthash, $headers = array()) {
      */
     public function define_client_check() {
         global $OUTPUT;
+
+        // TODO use MDL check api admin setting if available ???
```
matthewhilton, line 95: TODO use https://tracker.moodle.org/browse/MDL-67898 if available, since it uses Ajax so page loading times are very quick

classes/local/store/object_client_base.php
```diff
@@ -135,20 +152,20 @@ public function proxy_range_request(\stored_file $file, $ranges) {
      * Test proxy range request.
      *
      * @param  object  $filesystem  Filesystem to be tested.
-     * @return object
+     * @return result
      */
-    public function test_range_request($filesystem) {
-        return (object)['result' => false, 'error' => ''];
+    public function test_range_request($filesystem): result {
+        return new result(result::UNKNOWN, '');
```
matthewhilton: Maybe these should be `N/A` if not implemented? Otherwise things that dont implement them immediately get a `UNKNOWN` which is considered very bad.

classes/local/store/object_client_base.php
```diff
@@ -135,20 +152,20 @@ public function proxy_range_request(\stored_file $file, $ranges) {
...
      * Tests connection to external storage.
      * Override this method in client class.
      *
-     * @return object
+     * @return result
      */
-    public function test_connection() {
-        return (object)['success' => false, 'details' => ''];
+    public function test_connection(): result {
+        return new result(result::UNKNOWN, '');
```
matthewhilton: See above N/A instead of unknown

classes/local/store/object_client_base.php
```diff
@@ -158,7 +175,17 @@ public function test_connection() {
      * @param bool $testdelete Test delete permission and fail the test if could delete object from the storage.
      * @return object
      */
-    public function test_permissions($testdelete) {
-        return (object)['success' => false, 'details' => ''];
+    public function test_permissions($testdelete): result {
+        return new result(result::UNKNOWN, '');
```
matthewhilton: See above N/A instead of unknown

classes/local/store/object_client_base.php
```diff
@@ -158,7 +175,17 @@ public function test_connection() {
...
+    }
+
+    /**
+     * Tests configuration is OK.
+     * Override this method in client class.
+     *
+     * @return result
+     */
+    public function test_configuration(): result {
+        return new result(result::UNKNOWN, '');
```
matthewhilton: See above N/A instead of unknown

classes/local/store/s3/client.php
```diff
@@ -89,29 +93,56 @@ public function __wakeup() {
...
-        return true;
+        return new result($status, $summary, $details);
+    }
+
+    /**
+     * Checks the configuration is valid.
+     * @return array key => value where the value is true if ok, false if bad, or null if n/a
+     */
+    protected function check_configuration($config) {
+        return [
```
matthewhilton: Should this check more than just these - i.e. maybe the credentials check in here as well ?

classes/local/store/s3/client.php
```diff
@@ -244,100 +275,102 @@ protected function get_filepath_from_hash($contenthash) {
...
         } catch (\GuzzleHttp\Exception\InvalidArgumentException $e) {
-            $connection->success = false;
-            $connection->details = $this->get_exception_details($e);
+            $status = result::ERROR;
+            $details = $this->get_exception_details($e);
         } catch (\Aws\Exception\CredentialsException $e) {
-            $connection->success = false;
-            $connection->details = $this->get_exception_details($e);
+            $status = result::ERROR;
+            $details = $this->get_exception_details($e);
```
matthewhilton: For each of these just straight return a error status, don't use variables.

classes/local/store/s3/client.php
```diff
@@ -244,100 +275,102 @@ protected function get_filepath_from_hash($contenthash) {
...
         }
 
-        return $connection;
+        $summarystr = $status == result::OK ? 'check:passed' : 'check:failed';
+        $summary = get_string($summarystr, 'tool_objectfs');
+        return new result($status, $summary, $details);
```
matthewhilton: Per above, just return an OK result here and remove the variables

classes/local/store/s3/client.php
```diff
@@ -244,100 +275,102 @@ protected function get_filepath_from_hash($contenthash) {
...
                 // Something else went wrong.
                 if ($errorcode !== 'AccessDenied') {
                     $details = $this->get_exception_details($e);
-                    $permissions->messages[get_string('settings:deleteerror', 'tool_objectfs') . $details] = 'notifyproblem';
-                    $permissions->success = false;
+                    $details .= get_string('settings:deleteerror', 'tool_objectfs') . $details;
+                    $status = result::ERROR;
                 }
             }
         }
```
matthewhilton: Ideally revert this back to previous and just implode the array of messages at the end - a bit simpler and less diff

classes/local/store/swift/client.php
```diff
@@ -209,22 +206,20 @@ public function test_connection() {
...
-                $details = $this->get_exception_details($e);
-                $connection->messages[get_string('settings:connectionreadfailure', 'tool_objectfs') . $details] = 'notifyproblem';
-                $connection->success = false;
+                $details = get_string('settings:connectionreadfailure', 'tool_objectfs') . $this->get_exception_details($e);
+                return new result(result::ERROR, get_string('check:failed', 'tool_objectfs'), $details);
             }
         }
 
-        return $connection;
+        return new result(result::OK, get_string('check:success', 'tool_objectfs'));
```
matthewhilton: Lang string should be check:passed not check:success

classes/local/store/swift/client.php
```diff
@@ -209,22 +206,20 @@ public function test_connection() {
...
     }
 
-    public function test_permissions($testdelete) {
+    public function test_permissions($testdelete): result {
```
matthewhilton, line 227: PHP doc required

lib.php
```diff
@@ -117,11 +117,15 @@ function tool_objectfs_pluginfile($course, $cm, context $context, $filearea, arr
  * @return array
  */
 function tool_objectfs_status_checks() {
+    $checks = [
+        new tool_objectfs\check\configuration(),
+        new tool_objectfs\check\store(\tool_objectfs\check\store::TYPE_PERMISSIONS),
+        new tool_objectfs\check\store(\tool_objectfs\check\store::TYPE_CONNECTION),
+    ];
+
     if (get_config('tool_objectfs', 'proxyrangerequests')) {
```
matthewhilton, line 126: Should we check signing support here as well ? That's checked inside the proxy_range_request check

classes/local/store/object_client_base.php
matthewhilton: Changed to `INFO` which seems the most fitting https://moodledev.io/docs/apis/subsystems/check#result-states-of-a-check

matthewhilton:
Closing as this is only about 50% done and is a bit out of date with the newer branches

Issue #596, opened 2024-02-13
generate_status_report performance - SQL improvements


JoshAudette:
Issue:
The `\tool_objectfs\task\generate_status_report` task can be quite slow and costly in terms of database resources. 

Recently we have experienced SQL queries failing on MySQL with a message indicating temporary storage space full.  On reconfiguring MySQL to favour memory (`temptable_max_ram` tweak) we saw the query succeed for awhile, but lately the MySQL database exhausts memory causing it to restart.

We have had a look at changing the SQL queries used by the report.

See also: https://github.com/catalyst/moodle-tool_objectfs/issues/572

JoshAudette:
Hi,

We've come up with a SQL query using Common Table Elements approach.  This was meant to help make the queries easier to read / understand but in practice trials these perform much faster.  As well, the MySQL server doesn't exhaust resources running these, even when using a lower instance class.

Downside, MySQL 5.7 supposedly doesn't support CTE / the WITH clause, so I guess this would require a new branch for Moodle 4.2 where MySQL 8 is the minimum.

For location = DUPLICATED, EXTERNAL, ERROR
```
WITH
  cte_objects AS (
    SELECT 
    o.contenthash, 
    o.location
    FROM {tool_objectfs_objects} o
    WHERE o.location = ? ),
  cte_obj_files AS (
    SELECT 
    f.contenthash, 
    MAX(f.filesize) AS filesize
    FROM {files} f
    INNER JOIN cte_objects co ON f.contenthash = co.contenthash
    WHERE filesize > 0 
    GROUP BY f.contenthash, f.filesize )
SELECT
COALESCE(COUNT(cof.contenthash),0) AS objectcount,
COALESCE(SUM(cof.filesize),0) AS objectsum
FROM cte_obj_files cof
```

For location = LOCAL
```
WITH
  cte_objects AS (
    SELECT 
    o.contenthash, 
    o.location
    FROM {tool_objectfs_objects} o ),
  cte_obj_files AS (
    SELECT 
    f.contenthash, 
    MAX(f.filesize) AS filesize
    FROM {files} f
    LEFT JOIN cte_objects co ON f.contenthash = co.contenthash
    WHERE filesize > 0 AND ( co.location = ? OR co.location IS NULL )
    GROUP BY f.contenthash, f.filesize )
SELECT
COALESCE(COUNT(cof.contenthash),0) AS objectcount,
COALESCE(SUM(cof.filesize),0) AS objectsum
[...]

PR #597, opened 2024-02-15
Issue #596: SQL change with some performance / resource improvements

JoshAudette:
#596 

Updated SQL for the generate_status_report task using Common Table Elements.

In testing, these queries use less resources and complete faster compared to current ones.

Tested on MySQL 8 and PosgreSQL 14.

NOTE: the `WITH` clause is supported in MySQL since version 8 but not 5.7.  Since 12 in PostgreSQL.

So I think we may need a base branch MOODLE_402_STABLE and then change this PR to merge to that base branch.

### DIFF
--- a/classes/local/report/location_report_builder.php
+++ b/classes/local/report/location_report_builder.php
@@ -50,31 +50,52 @@ class location_report_builder extends objectfs_report_builder {
         $filedircount = 0;
         $filedirsum = 0;
         foreach ($locations as $location) {
-            $localsql = '';
-            if ($location == OBJECT_LOCATION_LOCAL) {
-                $localsql = ' or o.location IS NULL';
-            }
+            $sql =
+                'WITH
+                  cte_objects AS (
+                    SELECT o.contenthash, o.location
+                      FROM {tool_objectfs_objects} o
+                     WHERE o.location = ? ),
+                  cte_obj_files AS (
+                    SELECT f.contenthash, MAX(f.filesize) AS filesize
+                      FROM {files} f
+                INNER JOIN cte_objects co ON f.contenthash = co.contenthash
+                     WHERE filesize > 0
+                  GROUP BY f.contenthash, f.filesize)
+               SELECT COALESCE(COUNT(cof.contenthash),0) AS objectcount,
+                      COALESCE(SUM(cof.filesize),0) AS objectsum
+                 FROM cte_obj_files cof';
 
-            $sql = 'SELECT COALESCE(count(sub.contenthash) ,0) AS objectcount,
-                           COALESCE(SUM(sub.filesize) ,0) AS objectsum
-                      FROM (SELECT f.contenthash, MAX(f.filesize) AS filesize
-                              FROM {files} f
-                              LEFT JOIN {tool_objectfs_objects} o on f.contenthash = o.contenthash
-                              GROUP BY f.contenthash, f.filesize, o.location
-                              HAVING o.location = ?' . $localsql .') AS sub
-                     WHERE sub.filesize > 0';
+            if ($location == OBJECT_LOCATION_LOCAL) {
+                $sql =
+                    'WITH
+                      cte_objects AS (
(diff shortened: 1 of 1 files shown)


classes/local/report/location_report_builder.php
```diff
@@ -50,31 +50,52 @@ public function build_report($reportid) {
...
+            $sql =
+                'WITH
+                  cte_objects AS (
+                    SELECT o.contenthash, o.location
+                      FROM {tool_objectfs_objects} o
+                     WHERE o.location = ? ),
+                  cte_obj_files AS (
+                    SELECT f.contenthash, MAX(f.filesize) AS filesize
+                      FROM {files} f
+                INNER JOIN cte_objects co ON f.contenthash = co.contenthash
```
michaelkotlyar, line 62: Why is this an inner join as opposed to it previously being a left join?

classes/local/report/location_report_builder.php
```diff
@@ -50,31 +50,52 @@ public function build_report($reportid) {
...
+                        SELECT o.contenthash,  o.location
+                          FROM {tool_objectfs_objects} o ),
+                      cte_obj_files AS (
+                        SELECT f.contenthash, MAX(f.filesize) AS filesize
+                          FROM {files} f
+                     LEFT JOIN cte_objects co ON f.contenthash = co.contenthash
+                         WHERE filesize > 0 AND ( co.location = ? OR co.location IS NULL )
+                      GROUP BY f.contenthash, f.filesize)
+                   SELECT COALESCE(COUNT(cof.contenthash),0) AS objectcount,
+                          COALESCE(SUM(cof.filesize),0) AS objectsum
```
michaelkotlyar, line 82: These SQL strings are largely similar to each other. Any way we can split them up into reusable chunks?

michaelkotlyar:
Hey Josh, this is looking good. You'll see I've added some comments to resolve but once we've sorted that out, I'll create a 4.2 branch and merge this

classes/local/report/location_report_builder.php
JoshAudette: Only for case of location=0 / files that are LOCAL, the report counts not just files `where location=0` but also files that are only in {files} and not yet in {tool_objectfs_objects}, i.e. `where location is null`.  So in that case we need to `left join` so that the count includes files which aren't yet known to ObjectFS.

For the other cases e.g. DUPLICATED, EXTERNAL and ERROR, the report counts objects in both tables so we can use `INNER JOIN` and we can also move the `where location = ?` clause to reduce the `cte_objects` temp table size.  This really helps optimize the query.

We wanted to ensure these optimized queries produce the exact same numbers as the originals did.

classes/local/report/location_report_builder.php
JoshAudette: It could probably be done.  I guess there's a balance between reusable chunks versus being more readable.

The original query was both reusable and readable, with the "OR LOCATION IS NULL" string added conditionally to the query.  In the updated version that `WHERE` clause also shifts between temp tables depending if we do the `LEFT JOIN` or `INNER JOIN` version.

I think if we tried to split it up into more reusable chunks it may be harder to understand the code and the reasoning why the query changes.  We could make it even more readable by doing a switch-case since it's really those 3 cases where the query is different: 

- for LOCAL we use the LEFT JOIN, and the WHERE LOCATION = 0 OR LOCATION = NULL clause comes after the LEFT JOIN;
- for DUPLICATED, EXTERNAL and ERROR we can do the INNER JOIN and we shift the WHERE LOCATION = ? clause to the `cte_objects` temp table, both improving performance;
- for ORPHANED we're only counting objects from `{tool_objectfs_objects}`.

I think it's okay as it is, but we could have a go at splitting up into more reusable chunks if you want?

michaelkotlyar:
Hey Josh, all merged to our new [MOODLE_402_STABLE](https://github.com/catalyst/moodle-tool_objectfs/tree/MOODLE_402_STABLE) branch. Thanks for your help!

PR #603, opened 2024-03-01
Update README.md

phette23:
Using Google Cloud storage also requires the AWS SDK. This is perhaps obvious to people who know how Google's storage is AWS compatible but someone unfamiliar might miss this step when trying to use GCP.


danmarsden:
thanks for the PR! :-)

I note that Moodle core is pulling in the aws lib so we may not need this in future releases.

Issue #614, opened 2024-06-05
Add a ACL option to set access permission in S3 bucket


PhucNguyen0311:
There was an issue when we used CloudFront in one account to access S3 in another account. For example, my application and CloudFront are stored in A account but my S3 bucket is configured in B account. Then, I checked the log of scheduled tasks and I can see this issue:

php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\push_objects_to_storage'
Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 12:42:52. Current memory use 2.9┬áMB.
copy_object_from_local_to_external. Total time taken: 0.16533708572388 seconds. Location change summary:
**local -> local.** Objects moved: 4. Total size: 4.6┬áMB.
... used 13 dbqueries
... used 0.66407299041748 seconds
Scheduled task complete: Object file system upload task (tool_objectfs\task\push_objects_to_storage)

It means we can't push objects from local to s3 (local->external). I tried to add a debug in upload_to_s3() function in https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/s3/client.php line 458 and get this error:

Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 12:27:04. Current memory use 2.9┬áMB.
exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "PutObject" on "https://s3bucketname.s3.region.amazonaws.com/path/to/objectkey"; AWS HTTP error: Client error: `https://s3bucketname.s3.region.amazonaws.com/path/to/objectkey` resulted in a `400 Bad Request` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessControlListNotSupported</Code><Message>The bucket does not all (truncated...)
 AccessControlListNotSupported (client): **The bucket does not allow ACLs** - <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessControlListNotSupported</Code><Message>The bucket does not allow ACLs</Message><RequestId>RequestID</RequestId><HostId>HostID=</HostId></Error>'

[...]

PR #615, opened 2024-06-05
Add an ACL option to set access permission in S3 bucket #614

PhucNguyen0311:
There was an issue when we used CloudFront in one account to access S3 in another account. For example, my application and CloudFront are stored in A account but my S3 bucket is configured in B account. Then, I checked the log of scheduled tasks and I can see this issue:

php admin/cli/scheduled_task.php --execute='\tool_objectfs\task\push_objects_to_storage'
Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 12:42:52. Current memory use 2.9┬áMB.
copy_object_from_local_to_external. Total time taken: 0.16533708572388 seconds. Location change summary:
local -> local. Objects moved: 4. Total size: 4.6┬áMB.
... used 13 dbqueries
... used 0.66407299041748 seconds
Scheduled task complete: Object file system upload task (tool_objectfs\task\push_objects_to_storage)

It means we can't push objects from local to s3 (local->external). I tried to add a debug in upload_to_s3() function in https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_310_STABLE/classes/local/store/s3/client.php line 458 and get this error:

Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 12:27:04. Current memory use 2.9┬áMB.
exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "PutObject" on "https://s3bucketname.s3.region.amazonaws.com/path/to/objectkey"; AWS HTTP error: Client error: https://s3bucketname.s3.region.amazonaws.com/path/to/objectkey resulted in a 400 Bad Request response:

AccessControlListNotSupportedThe bucket does not all (truncated...)
AccessControlListNotSupported (client): The bucket does not allow ACLs -
AccessControlListNotSupportedThe bucket does not allow ACLsRequestIDHostID='

I recognized the $acl is set as private by default https://github.com/catalyst/moodle-local_aws/blob/master/sdk/Aws/S3/ObjectUploader.php line 54 but we don't have any way to input $acl from upload_to_s3() function in moodle-tool-objectfs plugin. I'm tested this and it doesn't matter if Moodle, CloudFront and S3 are stored in the same AWS account. So, I'm wondering if we can create an option for this in Moodle setting page with private by default.


bwalkerl:
Closing as this was merged to 4.2+ a while back as part of #661

PR #619, opened 2024-07-22
Add object tagging - Moodle 310

### LINKED ISSUE #11: Add setting to ignore backup files

### LINKED ISSUE #617: Consider adding s3 tags for file metadata
eg file area, file type etc to make it easier to debug

matthewhilton:
**Summary**
Allows objectfs to send tags for an object to external stores. Initially only S3 is supported. `location` and `environment` tags are added.

For technical reasons, `mimetype` is also stored but this is stored in metadata instead of the tags (and thus is only for new objects, migration of metadata is not feasible)

The use case for this is to be able to setup conditional lifecycle rules to send orphaned files to deep archive and eventually delete them.

See https://github.com/catalyst/moodle-tool_objectfs/issues/11#issuecomment-2235182086 for the rationale behind using S3 tags instead of other methods such as file prefixes or separate buckets. 

**How it works**
See TAGGING.md

**Cherry picks**
- Cherry picked to latest branch `MOODLE_402_STABLE` -> https://github.com/catalyst/moodle-tool_objectfs/pull/623

**Related issues**
- Closes #617 
- Closes #11


matthewhilton:
I've been pondering the idea of removing the intermediary table from the equation here.

Originally the idea is we cache the tag values in an intermediary table, so the values can be easily queried on the moodle side for reporting. This allows you to have 'expensive' tags i.e. those that are computationally expensive to calculate, but also be able to query them very easily using the DB for reporting.

However if we want to only have 'cheap' tags, we might be able to get away with just aggregating their values in the report data table (i.e every time report is run, it recalculates the values and sums the counts together), and then instead of querying db for the cached tags, it would simply re-calculate it the moment the object is about to get updated. 

The downside of this is the report generating gets a tad bit more complex and heavy because we need to constantly re-calculate the tag values (which can come from various random places in the db) as opposed to a single tag cache intermediary table.

Another downside is we wouldn't then be able to track how many have actually been tagged in the external store, whereas with an intermediary table we at least know if its in there, its value is 99% of the time the same in the external store.

I'm leaning towards the simpler solution which is just to have them in the intermediary table.

matthewhilton:
This is mostly ready to go, will give it a test for these scenarios:

**Manual testing**
- [x] New site install - enable tagging and confirm tags are uploaded when objects are replicated
- [x] Upgraded site with existing object - enable tagging and confirm tags are uploaded via migration and also when new objects are uploaded
- [x] Site without overwrite - confirm cannot overwrite objects and tags remain intact
- [x] Site with overwrite - confirm can overwrite and tags are overwritten and updated
- [x] Tagging enabled without permissions in S3 - Confirm exceptions are thrown correctly, and check API class alerts issue
- [x] Object report - generates object report with correct tag value counts.

classes/local/store/object_file_system.php
```diff
@@ -1154,4 +1163,51 @@ private function update_object(array $result): array {
...
+            throw new coding_exception("Cannot sync tags, external client does not support tagging.");
+        }
+
+        // Get a lock before syncing, to ensure other parts of objectfs are not moving/interacting with this object.
+        $lock = $this->acquire_object_lock($contenthash, 10);
+
+        // No lock - just skip it.
+        if (!$lock) {
+            throw new coding_exception("Could not get object lock");
+        }
```
bwalkerl: Is this something that's likely to occur or a rare edge case? Logically this is the safe approach and is probably fine, but if it's somewhat frequent we may need to look at doing something else.

classes/local/store/object_file_system.php
```diff
@@ -1154,4 +1163,51 @@ private function update_object(array $result): array {
...
+                tag_manager::store_tags_locally($contenthash, $tags);
+            }
+
+            // Regardless, it has synced.
+            tag_manager::mark_object_tag_sync_status($contenthash, tag_manager::SYNC_STATUS_SYNC_NOT_REQUIRED);
+        } catch (Throwable $e) {
+            $lock->release();
+
+            // Mark object as tag sync error, this should stop it re-trying until fixed manually.
+            tag_manager::mark_object_tag_sync_status($contenthash, tag_manager::SYNC_STATUS_ERROR);
```
bwalkerl, line 1226: Is there any way to recover from this besides manual SQL? I think requiring manual intervention is the right approach, but it might be worth having an inbuilt way to retry these via GUI or CLI.

classes/task/update_object_tags.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if (empty($iterationlimit) || empty($iteration)) {
+            mtrace("Invalid number of iterations, exiting.");
+            return;
+        }
+
+        if (abs($iteration) > abs($iterationlimit)) {
+            mtrace("Maximum number of iterations reached: " . $iteration . ", exiting.");
+            return;
+        }
```
bwalkerl: Is there a reason why abs is used? I can't see any harm, but it's interesting to see this used over invalidating negatives.

classes/local/store/s3/file_system.php
```diff
@@ -32,6 +32,7 @@
 
 use tool_objectfs\local\manager;
 use tool_objectfs\local\store\object_file_system;
+use tool_objectfs\local\tag\tag_manager;
```
bwalkerl: No other code changes in this file - is this needed?

classes/local/store/s3/client.php
```diff
@@ -875,4 +878,113 @@ public function test_range_request($filesystem) {
...
+    }
+
+    /**
+     * Returns given objects tags queried from the external store. Object must exist.
+     * @param string $contenthash file content has
+     * @return array array of key=>value tag pairs
+     */
+    public function get_object_tags(string $contenthash): array {
+        $key = $this->bucketkeyprefix . $this->get_filepath_from_hash($contenthash);
+
```
bwalkerl: Really minor, but the same key was named $objectkey in set_object_tags, so would be good to be consistent.

settings.php
```diff
@@ -23,6 +23,9 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+use tool_objectfs\local\tag\tag_manager;
+use tool_objectfs\task\update_object_tags;
```
bwalkerl, line 29: Is update_object_tags used?

settings.php
```diff
@@ -249,4 +251,42 @@
...
         new lang_string('settings:preferexternal', 'tool_objectfs'), '', ''));
+
+    // Tagging settings.
+    $settings->add(new admin_setting_heading('tool_objectfs/taggingsettings',
+        new lang_string('settings:taggingheader', 'tool_objectfs'), ''));
+
+    $settings->add(new admin_setting_description('tool_objectfs/tagginghelp',
+        '',
+        get_string('settings:tagging:help', 'tool_objectfs')
+    ));
```
bwalkerl: Could this be shown using the third param for admin_setting_heading ('information') instead?

tests/local/tagging_test.php
```diff
@@ -0,0 +1,340 @@
...
+                'enabledinconfig' => false,
+                'supportedbyfs' => false,
+                'expected' => false,
+            ],
+            'enabled in config but fs does not support' => [
+                'enabledinconfig' => true,
+                'supportedbyfs' => false,
+                'expected' => false,
+            ],
+            'enabled in config and fs does  support' => [
```
bwalkerl: Double space between 'does' and 'support'

lang/en/tool_objectfs.php
```diff
@@ -269,3 +269,22 @@
...
+$string['settings:maxtaggingiterations'] = 'Object tagging adhoc sync maximum number of iterations ';
+$string['settings:maxtaggingiterations:desc'] = 'The maximum number of times the tagging sync adhoc task will requeue itself. To avoid accidental infinite runaway.';
+$string['settings:overrideobjecttags'] = 'Allow object tag override';
+$string['settings:overrideobjecttags:desc'] = 'Allows ObjectFS to overwrite tags on objects that already exist in the external store.';
+$string['tagsource:environment'] = 'Environment defined by $CFG->objectfs_environment_name, currently: "{$a}".';
+$string['tagsource:mimetype'] = 'File mimetype as stored in {files} table';
+$string['settings:tagsources'] = 'Tag sources';
+$string['task:triggerupdateobjecttags'] = 'Queue adhoc task to update object tags';
+$string['settings:tagging:help'] = 'Object tagging allows extra metadata to be attached to objects in the external store. Please read TAGGING.md in the plugin Github repository for detailed setup and considerations. This is currently only supported by the S3 external client.';
+$string['object_status:tag_count'] = 'Object tags';
```
bwalkerl: The ordering here looks to be in a rough temporary state. I would at least expect some basic ordering within this section.

classes/local/tag/tag_manager.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Stores tag records for contenthash locally
+     * @param string $contenthash
+     * @param array $tags
+     */
+    public static function store_tags_locally(string $contenthash, array $tags) {
+        global $DB;
+
+        // Purge any existing tags for this object.
+        $DB->delete_records('tool_objectfs_object_tags', ['contenthash' => $contenthash]);
```
bwalkerl: Deleting and inserting here feels a bit off to me, especially when done at scale for future migrations. Is there any reason why this was chosen over update/inserting aside from simplicity?

classes/local/tag/tag_manager.php
```diff
@@ -0,0 +1,183 @@
...
+        $recordstostore = [];
+        foreach ($tags as $key => $value) {
+            $recordstostore[] = [
+                'contenthash' => $contenthash,
+                'tagkey' => $key,
+                'tagvalue' => $value,
+                'timemodified' => $timemodified,
+            ];
+        }
+        $DB->insert_records('tool_objectfs_object_tags', $recordstostore);
```
bwalkerl: One concern is that if a new tag is added in the future the timemodified of the old tags will be updated. This may be technically correct as PutObjectTagging does update all, but would limiting this to when it was actually modified be more useful?

db/install.xml
```diff
@@ -49,5 +50,20 @@
...
+        <FIELD NAME="tagkey" TYPE="char" LENGTH="32" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="tagvalue" TYPE="char" LENGTH="128" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="timemodified" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="false"/>
+      </FIELDS>
+      <KEYS>
+        <KEY NAME="primary" TYPE="primary" FIELDS="id"/>
+      </KEYS>
+      <INDEXES>
+        <INDEX NAME="objecttagkey_idx" UNIQUE="true" FIELDS="contenthash, tagkey"/>
+      </INDEXES>
```
bwalkerl: This table could end up being quite large on sites with a lot of objects. I agree with your reasoning for keeping it on the PR, especially tracking how many sites have been tagged, but I think we should be careful with the size.

Could contenthash potentially be replaced with object id?

classes/local/report/objectfs_report.php
```diff
@@ -166,6 +166,7 @@ public static function get_report_types() {
             'location',
             'log_size',
             'mime_type',
+            'tag_count',
```
bwalkerl, line 170: If the tag count report has no rows, this will display "Nothing to display" at the bottom of the report. It may be worth trying to make this message more meaningful or hiding it under certain conditions i.e. if no rows and/or tags are disabled.

TAGGING.md
```diff
@@ -0,0 +1,86 @@
...
+Note the rules about sources:
+- Identifier must be < 32 chars long.
+- Value must be < 128 chars long.
+
+While external providers allow longer key/values, we intentionally limit it to reserve space for future use. These limits may change in the future as the feature matures.
+
+To add a new source:
+- Implement `tag_source`
+- Add to the `tag_manager` class
+- As part of an upgrade step, mark all objects `tagsyncstatus` to needing sync (using `tag_manager` class, or manually in the DB)
```
bwalkerl, line 64: Having a specific function in tag_manager to reset sync status for all objects would be useful for this.

classes/task/update_object_tags.php
bwalkerl: The iteration limit is a great fallback, but I think there needs to be more visibility for when the limit the reached instead of being hidden in the task logs. User intervention is most likely required as failing this would either mean there's an actual problem or that a migration hasn't been completed (in the case of a very large site that hasn't set appropriate limits).

bwalkerl (commented):
Hey @matthewhilton  this is looking great from the code perspective. I've pointed out a few things I've noticed, but these are mostly just some extra considerations and discussion points. I've focused more on reviewing the code and haven't tested this end to end myself - so I'm partly relying on your test results for that. 

My only real concern here is the perform at scale, mostly when back filling with the adhoc task, but this is an optional feature so I think it should be fine to roll out and test as needed.

classes/local/store/object_file_system.php
brendanheywood: I'd like to understand the failure mode a bit more here. Ideally if we can't get a lock then something else is doing stuff, and it might be doing the same stuff. So we could wait with a 0 timeout, and then just keep moving. The question is will this then get picked up on the next cron run and retried?

matthewhilton: It's an edge case. If this is hit by the migration task (where I think it will be hit the most) it will retry since the sync status will still be 'needs sync', but if this is hit in the course of normal file upload (where it really shouldn't ever) I think it will just remain untagged until a migration task is run.

I imagine a scenario you would most likely hit this is if you are uploading a really big file, and the record exists in mdl_files but objectfs is still uploading it, and the tagging migration task tries to tag it at the same time, but gets rejected because the original process still has the object lock.

classes/local/store/object_file_system.php
matthewhilton: I wasn't expecting this to get triggered often, it was more of a failsafe so we don't have infinitely retrying tags racking up tons of unnecessary api call costs.

But agreed I think maybe a simple way to see how many and retry them from the frontend could be good.

classes/local/tag/tag_manager.php
matthewhilton: No this was purely for simplicity. Its so much easier and performant to just do a bulk delete records and a bulk insert, as opposed to an update diff.

A diff also has problems if you ever change the key of a tag, it would unlink it and the old records would float around and need to be handled, as opposed to doing it this way which ensures the old state is always wiped away and replaced with the new state.

classes/local/tag/tag_manager.php
matthewhilton: Currently as it is setup all the tags for the same contenthash have the same timemodified. I left it this way because it is simpler to just add it here rather than what is probably more 'correct' which is having a single `timetagssynced` on the `tool_objectfs_objectfs` record that is updated.

I think I might remove the timemodified here and move it to the object record, unless we decide / work out if we are going to diff them or just reset the state each time (above comment) which then timemodified per tag record would make more sense.

classes/task/update_object_tags.php
matthewhilton: No reason you are right through it should throw an exception, will tweak

db/install.xml
matthewhilton: I went back and forth on this myself for a lot.

Contenthash is so much simpler in terms of DB operations (less lookups, etc...) and code, because we don't have to lookup the object id (since almost all the existing objectfs apis use contenthash, not object id)

But you are right, it means we have a long string instead of an id, hence it takes up more storage.

I lent towards simplicity, but i'll look at it again and see if it can be implemented nicely without too many extra db operations.

matthewhilton:
I've not addressed all the comments yet, but just an fyi pushed an update to add a bit more visibility to the migration progress, now there is a nice table in the admin settings (for a lack of a better place to put it), let me know if you see any tweaks that should be made here, or if there are better styling options available (I couldn't find any):

image

Additionally pushed a CI fix, since moodle plugin ci looks to have updated on my local and was throwing errors where it wasn't before. Note the Github CI is failing due to an upstream nvm issue which i've been tracking but is not fixed yet (so it will stay red for now)

settings.php
```diff
@@ -249,4 +251,54 @@
...
+
+    // Tagging status.
+    $settings->add(new admin_setting_heading('tool_objectfs/taggingstatus',
+        new lang_string('settings:taggingstatus', 'tool_objectfs'), ''));
+
+    // Build overview of tag statuses.
+    $settings->add(new admin_setting_description('tool_objectfs/taggingstatuscounts',
+        new lang_string('settings:taggingstatuscounts', 'tool_objectfs'),
+        tag_manager::get_tag_sync_status_summary_html()
+    ));
```
bwalkerl: I wouldn't recommend having this in the main admin settings as the query will get called on other site admin pages. If you can't find a suitable location maybe an external admin page would work?

classes/task/update_object_tags.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Generates a html table summarising the update object tag tasks that exist
+     * @return string
+     */
+    public static function get_summary_html(): string {
+        $tasks = manager::get_adhoc_tasks(self::class);
+
+        if (empty($tasks)) {
+            return get_string('tagging:migration:nothingrunning');
```
bwalkerl: Missing component.

classes/local/store/object_file_system.php
```diff
@@ -1154,4 +1163,53 @@ private function update_object(array $result): array {
...
+
+            if ($canset) {
+                $tags = tag_manager::gather_object_tags_for_upload($contenthash);
+                $this->get_external_client()->set_object_tags($contenthash, $tags);
+                tag_manager::store_tags_locally($contenthash, $tags);
+                tag_manager::record_tag_pushed_time($contenthash, time());
+            }
+
+            // Regardless, it has synced.
+            tag_manager::mark_object_tag_sync_status($contenthash, tag_manager::SYNC_STATUS_COMPLETE);
```
bwalkerl: I think this would do 2 separate db updates on the same table. Perhaps it would be better to combine record_tag_pushed_time() with mark_object_tag_sync_status() by adding an optional param for time?

classes/task/update_object_tags.php
```diff
@@ -0,0 +1,150 @@
...
+            throw new moodle_exception('tagging:migration:limitreached', 'tool_objectfs', '', $iteration);
+        }
+
+        $fs = get_file_storage()->get_file_system();
+
+        // This is checked above in tag_manager::is_tagging_enabled_and_supported, but as a sanity check
+        // ensure this specific method is defined.
+        if (!method_exists($fs, "push_object_tags")) {
+            throw new coding_exception("Filesystem does not define push_object_tags");
+        }
```
bwalkerl, line 99: I don't think it's possible to recover from these exceptions, so when they are retried they are just going to keep increasing the fail delay. This is expected, but Moodle 4.4 introduced a few changes to ad hoc tasks that might be suitable to implement here - one example is the no-retry flag from MDL-79128 if it's backwards compatible with the 4.2 branch. It'd also be worth checking if any changes are needed to the check on get_fail_delay() because of this.

bwalkerl (commented):
Thanks @matthewhilton - feels like you went above and beyond with the changes so far. I've left a couple of new comments on these.

classes/local/tag/tag_manager.php
matthewhilton: I've changed it over now so the single object record has the `tagpushedtime`, instead of each tag record individually. I don't think there is enough value in having each have a timemodified to justify the required extra complexity of calculating and storing the diff (maybe in the future as a separate pr)

classes/local/report/objectfs_report.php
matthewhilton: This is kinda baked into the flexible table itself and not something you are meant to be able to change - see https://github.com/moodle/moodle/blob/6751ba650055ec1146df5e05b12972fa234b37f8/lib/tablelib.php#L1025

To make it a tiny bit better I've added the header for each table above it - it's a bit redundant for the others (since they have the name of the table in the table anyway) but it's the 'least bad' solution i'm aware of:
image

TAGGING.md
matthewhilton: I kinda want to lean towards not implementing methods unless there is an actual for them, since currently such a method would not be called anywhere.

I don't think at this stage manually updating the DB is that difficult, especially since it is explained in the readme what to do.

settings.php
matthewhilton: Fair call, I decided to move it to a check api class, and then just link them from the admin settings, which has the added benefit of if they warn about failures, they will alert the admins (e.g. via heartbeat)

It would be super nice if https://tracker.moodle.org/browse/MDL-67898 was available in lower versions so we could put the check straight into the settings but since its 4.4+ and that is somewhat far away from the supported 3.10 branch I decided to against it (e.g. gated behind `class_exists`)

classes/local/store/object_file_system.php
matthewhilton: Correct, it was just simpler (code-wise) that way. But since you mentioned it I looked at it again and managed to make it do one update call with a tiny bit of extra code complexity. :+1:

classes/local/report/objectfs_report.php
bwalkerl: Thanks Matt, the header resolves the main issue of the confusing message. I was thinking it may be possible to add some logic in get_report_types() to decide whether you add 'tag_count', but the header change should be enough.

TAGGING.md
bwalkerl: Completely understandable. Part of the reason I brought this up was to improve the visibility since I missed reading it in the first place, but that'd be entirely on the user for not reading the instructions properly.

db/install.xml
bwalkerl: Realistically this isn't too much of a concern at the current time with only 2 classes of tags, so I'm happy to go with whatever you think is best.

classes/local/tag/tag_manager.php
bwalkerl: If the bulk delete and insert is actually more performant I'm happy to leave it as is for now, since as you mention it does properly cover all scenarios.

It would be interesting to see some tests results around this, but we can always come back to this later if needed.

db/install.xml
matthewhilton: Had another look at it and really we only need the objectid when storing the tags, everything else can still be based off hash.

There were two options though to get the object id here:

1. Drill the objectid down from here https://github.com/catalyst/moodle-tool_objectfs/blob/d0f301b9bc4e2682119c81a434b236f2c860c655/classes/local/object_manipulator/pusher.php#L72 all the way to `push_object_tags`, which would save 1 db query but requires quite a bit of api changes just for this
2. Re-query it when saving tags just using `$DB->get_field`

I went with option 2 since its simpler code-wise and really 1 db query is not bad considering most objects will have that operation run once in its entire lifetime.

Lmk if you have different thoughts about this though

classes/task/update_object_tags.php
matthewhilton: We still want this to retry for e.g. if there was a blib in the connectivity and the s3 api failed for some reason (but then came good later). Not sure if its worth it to try micro optimise this, I think its ok if it fails (even if unrecoverable) since the intention is its a migration process that is watched actively not something triggered automatically.

Also checked and the fail delay remains unchanged with `MDL-79128` - the faildelay will still increase it will just stop trying to execute it after the `attemptsavailable` are exhausted.

db/install.xml
bwalkerl: I agree with option 2 if this is the only feature that would benefit from 1.

matthewhilton:
Just marking as draft temporarily, will be adding some features we discussed internally and then will re-open it.

matthewhilton:
Had a bit of time to work on object orphaning support.

The big issue i've found so far is that when an object is orphaned, we lose its mimetype. This is because when it goes to update the object tags after it is orphaned, the mimetype data in the mdl_files table no longer exists.

There are couple of ways to go about this:
1. Not care about mimetype anymore. Do we really need it ?
2. Push mimetype as metadata, instead of tags. This can only be done on upload (i.e. only new objects) but it is immutable so regardless of what we do with tags, the mimetype still remains. Theres no extra costs for metadata
3. Instead of the current 'replace all' tags on update, we implement some complex kind of reconciliation mechanism and only update the tags necessary. This probably has loads of edge cases.

IMO I think option 2 is the most balanced approach

matthewhilton:
I've finished the support for tagging orphaned objects.

We now have a simple `location` tag which has values `orphaned` or `active`.

I split this up from the env tag since the env tag is now also used by objectfs to determine if it should overwrite tags (mainly used for lower envs that we don't want messing with prod shared objects)

I also had to move mimetype from the tags to the metadata for [reasons mentioned above](https://github.com/catalyst/moodle-tool_objectfs/pull/619#issuecomment-2306384385).

classes/local/report/tag_count_report_builder.php
```diff
@@ -0,0 +1,51 @@
...
+        global $DB;
+        $report = new objectfs_report('tag_count', $reportid);
+
+        // Returns counts + sizes of key:value.
+        $sql = "
+            SELECT CONCAT(COALESCE(object_tags.tagkey, '(untagged)'), ': ', COALESCE(object_tags.tagvalue, '')) as datakey,
+                   COUNT(objects.id) as objectcount,
+                   SUM(objects.filesize) as objectsum
+              FROM {tool_objectfs_objects} objects
+         LEFT JOIN {tool_objectfs_object_tags} object_tags
```
brendanheywood: I'm quite anxious about this left join at scale

classes/local/store/object_file_system.php
brendanheywood: If you can't get a lock this should be business as usual, this should not be a coding_exception which by definition implies the code needs to be fixed

I don't think this should be an exception at all as it would break the file interaction, we want to just ignore it and move on and assume that it will get picked up again later by cron?

settings.php
```diff
@@ -249,4 +251,52 @@
...
+        new lang_string('settings:overrideobjecttags', 'tool_objectfs'),
+        get_string('settings:overrideobjecttags:desc', 'tool_objectfs'),
+        1
+    ));
+
+    // Tagging status.
+    $settings->add(new admin_setting_heading('tool_objectfs/taggingstatus',
+        new lang_string('settings:taggingstatus', 'tool_objectfs'), ''));
+
+    // TODO when this branch supports 4.4+, use admin_setting_check https://tracker.moodle.org/browse/MDL-67898.
```
brendanheywood: Why not add this now conditionally?

classes/local/report/tag_count_report_builder.php
matthewhilton: This is in the report builder so it will be done via adhoc tasks and never hit via web. However saying that the reason its joining is to get the filesize which is currently expected by the objectfs report handlers, despite it not really being that useful in the context of the report.
I could probably tweak that API a bit so we don't need the filesize hence dont need the join will see what I can do

matthewhilton: Turns out it was really simple to remove the object size from the report and hence remove the need for the left join - have changed this in https://github.com/catalyst/moodle-tool_objectfs/pull/619/commits/3002355ed9c5e46669fee456416daf5b718ae0ff

settings.php
matthewhilton: good idea - added

classes/local/store/object_file_system.php
matthewhilton: Agreed i've made it just return a bool instead of throwing an exception.

Thinking about it more, it would only ever get hit if you are uploading new objects and the migration cron task is also running and they happen to be trying to push tags for the same object at the same time.

If this did happen, regardless of who wins getting the lock, the object will still get tagged in the end so not a big deal.

matthewhilton:
Merging this one as it has been thoroughly reviewed and tested. Will squash into a single commit and cherry pick over to the other stables.

PR #621, opened 2024-07-29
Fix PHP 8.2 deprecation warnings in unit tests (4.2)

sammarshallou:
There were a number of 'Creation of dynamic property...' deprecation warnings when running the unit test. These happen if you set '$this->field = value' without declaring e.g. 'protected $field' in the class.

I've fixed these and the unit tests now run for me. The fix should be safe for all recent PHP versions, i.e. I didn't add types to the declarations.

This is just a MOODLE_402 branch version of #586.


danmarsden:
Thanks Sam!

PR #622, opened 2024-07-29
Fix codestandards 402

matthewhilton:
**Changes**
- A port of https://github.com/catalyst/moodle-tool_objectfs/pull/620 to `MOODLE_402_STABLE` which fixes things with `phpcbf` as well as manually fixing unit test data providers
- Also copied over a commit made to 310 but not 402 - 00b514ffcc8da92ca8f1bb71624b382862a023f0 to fix phpdocs as well

### DIFF
--- a/classes/local/object_manipulator/candidates/candidates_factory.php
+++ b/classes/local/object_manipulator/candidates/candidates_factory.php
@@ -33,6 +33,9 @@ use tool_objectfs\local\object_manipulator\pusher;
 use tool_objectfs\local\object_manipulator\recoverer;
 use tool_objectfs\local\object_manipulator\orphaner;
 
+/**
+ * Candidates Factory
+ */
 class candidates_factory {
 
     /** @var array $manipulatormap */
@@ -46,8 +49,10 @@ class candidates_factory {
     ];
 
     /**
-     * @param $manipulator
+     * Finder
+     * @param mixed $manipulator
      * @param stdClass $config
+     *
      * @return mixed
      * @throws moodle_exception
      */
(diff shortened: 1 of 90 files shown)


classes/local/object_manipulator/candidates/candidates_factory.php
```diff
@@ -33,6 +33,9 @@
 use tool_objectfs\local\object_manipulator\recoverer;
 use tool_objectfs\local\object_manipulator\orphaner;
 
+/**
+ * Candiates Factory
```
danmarsden: is this a factory where candy is eaten or just a typo? :-)

classes/local/object_manipulator/candidates/candidates_finder.php
```diff
@@ -27,6 +27,9 @@
 use moodle_exception;
 use stdClass;
 
+/**
+ * Candiates Finder
```
danmarsden: another one.

classes/privacy/provider.php
```diff
@@ -36,7 +36,7 @@ class provider implements null_provider {
      *
      * @return  string
      */
-    public static function _get_reason() {
+    public static function get_reason(): string {
```
danmarsden, line 37: this looks like a conversion from using the old legacy polyfill, but the use legacy_polyfill is still in place?

classes/local/object_manipulator/candidates/candidates_factory.php
matthewhilton: Just a typo :)

classes/local/object_manipulator/candidates/candidates_finder.php
matthewhilton: Fixed

classes/privacy/provider.php
matthewhilton: Ah yep, good pickup. Wasn't aware of the polyfill, and have removed since its no longer necessary.

Issue #634, opened 2024-09-16
Local file returns as `is_file_readable_externally_by_hash() == true` if `preferexternal` is set and file is 0 bytes in size


petersistrom:
If a local file has a size of 0 bytes and the `tool_objectfs | preferexternal` setting is turned on, then the `is_file_readable_externally_by_hash` https://github.com/catalyst/moodle-tool_objectfs/blob/6388430d84de4840d749ae746f94c515abcfc269/classes/local/store/object_file_system.php#L249 will still return true and the plugin will look for a non-existent external file and return a 404, even though the file exists locally.

petersistrom:
Related to #609

PR #638, opened 2024-09-18
Issues 633 634-MOODLE_402_STABLE

### LINKED ISSUE #633: Preferexternal setting is under `test settings`
image
This setting doesn't feel like a test setting and seems it should be moved to different section, potentially `General Settings`

petersistrom:
Resolves #633 and #634

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -246,13 +246,13 @@ abstract class object_file_system extends \file_system_filedir {
      * @return bool
      */
     public function is_file_readable_externally_by_hash($contenthash) {
-        if ($contenthash === sha1('')) {
-            // Files with empty size are either directories or empty.
+        $path = $this->get_external_path_from_hash($contenthash, false);
+
+        if ($contenthash === sha1('') && is_dir($path)) {
+            // Explicitly check if it is a directory, handle empty files as normal.
             // We handle these virtually.
             return true;
         }
-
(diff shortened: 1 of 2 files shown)


settings.php
```diff
@@ -110,6 +110,9 @@
         new lang_string('settings:tokenexpirywarnperiod', 'tool_objectfs'),
         '', 60 * DAYSECS, DAYSECS));
 
+    $settings->add(new admin_setting_configcheckbox('tool_objectfs/preferexternal',
```
brendanheywood, line 113: why have the settings changed?

classes/local/store/object_file_system.php
```diff
@@ -246,14 +246,14 @@ public function is_file_readable_externally_by_storedfile(stored_file $file) {
...
      */
     public function is_file_readable_externally_by_hash($contenthash) {
-        if ($contenthash === sha1('')) {
-            // Files with empty size are either directories or empty.
+        $path = $this->get_external_path_from_hash($contenthash, false);
+
+        if ($contenthash === sha1('') && is_dir($path)) {
+            // Explicitly check if it is a directory, handle empty files as normal.
             // We handle these virtually.
             return true;
```
brendanheywood, line 254: wouldn't it be better to simply say no this file isn't readable externally, so false here? get_external_path_from_hash is more expensive so I think this should come after the sha1 test like it was

petersistrom:
Created new pr #639

PR #641, opened 2024-09-19
Issue #633: Move preferexternal setting

petersistrom:


settings.php
```diff
@@ -247,10 +250,4 @@
             }
         }
     }
-
-    $settings->add(new admin_setting_heading('tool_objectfs/testsettings',
```
brendanheywood, line 251: is this lang string no longer used now?

petersistrom: Ah yes it is no longer used, updated to remove the lang string

bwalkerl:
Merged to 4.4+ as part of #765 and not needed on older branches.

PR #642, opened 2024-09-19
Issue #633: Move preferexternal setting

### LINKED ISSUE #633: Preferexternal setting is under `test settings`

petersistrom:
Resolves #633


bwalkerl:
Merged to 4.4+ as part of #765 and not needed on older branches.

PR #646, opened 2024-10-24
bugfix: fix token expiry check in admin settings

matthewhilton:
**Issue:**
There were a few issues found after more thorough testing of  https://github.com/catalyst/moodle-tool_objectfs/pull/636:
1. The class wasn't included (oops!)
2. There is a core bug which means on newer php versions if it didn't have an action link it would throw an exception -> https://tracker.moodle.org/browse/MDL-83537
3. The admin_setting_check doesn't work unless it is in the top level settings page and not in the client settings definition area, because of the way internally it searches through the admin tree to find the check.

**Fixes:**
1. Moved token expiry check to top level. Realistically all client sdks could/should implement this so no point having it only in the Azure client section.
2. Added action link as an fix until MDL-83537 is implemented
3. Fixed a spelling error

**Testing/screenshots:**
Where token expiry is implemented - in Azure client code:

Where token expiry is not implemented:

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -23,6 +23,8 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+use tool_objectfs\check\token_expiry;
+
 defined('MOODLE_INTERNAL') || die();
 
 require_once(__DIR__ . '/classes/local/manager.php');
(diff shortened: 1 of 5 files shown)


settings.php
```diff
@@ -248,6 +250,15 @@
...
 
+    // Admin_setting_check only exists in 4.5+, in lower versions fallback to a basic description.
+    if (class_exists('admin_setting_check')) {
+        $settings->add(new admin_setting_check('tool_objectfs/check_tokenexpiry', new token_expiry(), true));
+    } else {
+        $summary = (new token_expiry())->get_result()->get_summary();
+        $settings->add(new admin_setting_description('tool_objectfs/tokenexpirycheckresult',
+            get_string('checktoken_expiry', 'tool_objectfs'), $summary));
+    }
+
```
rhell4, line 264: Should this maybe be moved up further or under it's own heading? Right now for example with s3 set up it's at the bottom of Pre-Signed URLs Settings, which seems a little confusing, to an unknowing user it might look like it's referring to a pre-signed url token.

matthewhilton: yeah that is a really good point about confusing it for the pre signed url token. Have moved it to its own heading :+1:

rhell4:
Looks good to me.

Issue #645, opened 2024-10-14
`$testdelete` setting in client does not respect plugin config, delete permissions checks may be outdated


matthewhilton:
In the client class, there is a flag `$testdelete`, which if true it tests deleting objects in `test_permissions` which is run/displayed in the admin settings.

However, it does not follow the `Delete external objects` setting, which means for example if I have `Delete external objects` set to `no`, it will still try and delete but fail because I didn't give it delete permissions.

In the Digitalocean client class, this is just plainly disabled for an unknown reason, see:
https://github.com/catalyst/moodle-tool_objectfs/blob/398cabbb656b97bd84b695e8e9dcffe44d1a08a4/classes/local/store/digitalocean/client.php#L43

Furthermore, some classes e.g. the azure storage class do not even consider the $testdelete flag at all, however mysteriously it does not appear to fail when not given delete permissions (perhaps the error code changed?):

https://github.com/catalyst/moodle-tool_objectfs/blob/398cabbb656b97bd84b695e8e9dcffe44d1a08a4/classes/local/store/azure/client.php#L285-L295

I believe a similar problem might exist with the S3 class, where it is not correctly testing the delete permissions so its always trying to test but even if it can't delete it doesn't trigger an error when it should.

PR #647, opened 2024-11-05
Updated Azure storage compatibility

matthewhilton:
**Changes**
- Adds new `azure_blob_storage_file_system` class to replace the deprecated `azure_file_storage_system`.
    - This is intentionally added as an additional class (as opposed to renaming/fixing the old class) because the new class requires a new plugin dependency https://github.com/catalyst/moodle-local_azureblobstorage 
    - Having a new class allows time for admins to install the new plugin in parallel, and then switch over rather than suddenly finding the newest minor version update changed dependencies.
- Added documentation about migration process for the above
- Fixes a tiny bug in a test that assumed no files existed before the test
- Fixed a bug where testing deletion does not check if the fs even is allowed to delete, related to https://github.com/catalyst/moodle-tool_objectfs/issues/645 but does not completely resolve the issue for other fs's like AWS

**Testing**
- The new client class with the new sdk plugin pass the integration tests (the standard objectfs unit tests but attached to azure storage via the `azure_blob_storage_file_system`)
- Also tested locally the various CRUD operations including large files uploaded via multipart upload and large files streamed down (e.g. videos) in multipart - all working as expected

### DIFF
--- a/classes/local/manager.php
+++ b/classes/local/manager.php
@@ -306,17 +306,18 @@ class manager {
     public static function get_available_fs_list() {
         $result[''] = get_string('pleaseselect', OBJECTFS_PLUGIN_NAME);
 
-        $filesystems['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system';
+        $filesystems['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system [DEPRECATED]';
         $filesystems['\tool_objectfs\digitalocean_file_system'] = '\tool_objectfs\digitalocean_file_system';
         $filesystems['\tool_objectfs\s3_file_system'] = '\tool_objectfs\s3_file_system';
         $filesystems['\tool_objectfs\swift_file_system'] = '\tool_objectfs\swift_file_system';
+        $filesystems['\tool_objectfs\azure_blob_storage_file_system'] = '\tool_objectfs\azure_blob_storage_file_system';
 
-        foreach ($filesystems as $filesystem) {
+        foreach ($filesystems as $filesystem => $name) {
             $clientclass = self::get_client_classname_from_fs($filesystem);
             $client = new $clientclass(null);
 
             if ($client && $client->get_availability()) {
-                $result[$filesystem] = $filesystem;
+                $result[$filesystem] = $name;
             }
         }
         return $result;
--- a/classes/local/store/azure_blob_storage/client.php
+++ b/classes/local/store/azure_blob_storage/client.php
@@ -92,7 +92,7 @@ class client extends object_client_base {
      */
     public function get_fullpath_from_hash($contenthash): string {
         $filepath = $this->get_filepath_from_hash($contenthash);
-        return "blob://$filepath";
+        return "azure://$filepath";
     }
 
     /**
@@ -113,7 +113,7 @@ class client extends object_client_base {
      * @return string
      */
     protected function get_blob_key_from_path(string $filepath): string {
-        return str_replace("blob://", '', $filepath);
+        return str_replace("azure://", '', $filepath);
     }
 
     /**
@@ -171,7 +171,7 @@ class client extends object_client_base {
      */
     public function get_seekable_stream_context() {
         $context = stream_context_create([
-            'blob' => [
+            'azure' => [
                 'seekable' => true,
             ],
         ]);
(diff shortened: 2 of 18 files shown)


classes/local/store/azure_blob_storage/client.php
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Sets the StreamWrapper to allow accessing the remote content via a blob:// path.
+     */
+    public function register_stream_wrapper() {
+        if ($this->get_availability()) {
+            stream_wrapper::register($this->api);
+        } else {
+            parent::register_stream_wrapper();
```
Peterburnett, line 84: This is just void in the parent, perhaps we just early return instead here, its a bit clearer that this actually does nothing

classes/local/store/azure_blob_storage/client.php
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Returns the filepath from the contenthash, mimicking the
+     * structure of the filedir storage system.
+     * @param string $contenthash
+     * @return string filepath
+     */
+    protected function get_filepath_from_hash($contenthash): string {
+        $l1 = $contenthash[0] . $contenthash[1];
+        $l2 = $contenthash[2] . $contenthash[3];
+        return "$l1/$l2/$contenthash";
```
Peterburnett, line 107: Do we need to support ObjectFS bucket object prefixing here?

classes/local/store/azure_blob_storage/client.php
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * Renames a given file
+     * @param string $currentpath
+     * @param string $destinationpath
+     */
+    public function rename_file($currentpath, $destinationpath) {
+        // Azure does not support renaming, instead the file is copied
+        // and the old one is deleted.
+        copy($currentpath, $destinationpath);
```
Peterburnett, line 136: We should check for success here before actually deleting the original object

classes/local/store/azure_blob_storage/client.php
```diff
@@ -0,0 +1,296 @@
...
+            'success' => true,
+            'details' => '',
+        ];
+    }
+
+    /**
+     * Returns token expiry time
+     * @return int
+     */
+    public function get_token_expiry_time(): int {
```
Peterburnett, line 254: A part of me wonders if this should be in the API layer

classes/local/manager.php
```diff
@@ -306,17 +306,18 @@ public static function check_file_storage_filesystem() {
     public static function get_available_fs_list() {
         $result[''] = get_string('pleaseselect', OBJECTFS_PLUGIN_NAME);
 
-        $filesystems['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system';
+        $filesystems['\tool_objectfs\azure_file_system'] = '\tool_objectfs\azure_file_system [DEPRECATED]';
```
Peterburnett, line 309: Just confirming that the value is only used for display, and isn't used as a key later on in selection?

Peterburnett:
Should we be emitting debugging() anywhere when the old SDK is used? It has the potentially to be EXTREMELY noisy, just wondering if it would be useful with out drowning out noise.

classes/local/store/azure_blob_storage/client.php
matthewhilton: could be either or - all the other clients use this same pattern where it calls the parent function. But on the other hand it does nothing (at the moment) so why call it?

I'm leaning towards just keeping it in sync with the others, and if we wanted to we should just remove it from all of them (in a different pr)

classes/local/store/azure_blob_storage/client.php
matthewhilton: just checked, bucket key prefix is a s3 client specific setting so doesn't apply here

classes/local/manager.php
matthewhilton: Yes that's what the change on 315 and 320 were for :+1:

matthewhilton:
> Should we be emitting debugging() anywhere when the old SDK is used? It has the potentially to be EXTREMELY noisy, just wondering if it would be useful with out drowning out noise.

The old SDK already emits enough debugging messages on its own because of the php deprecation notices. I don't think we need to make it any worse and believe people will have enough reason to switch over when it becomes apparent the old one simply doesn't work anymore.

classes/local/store/azure_blob_storage/client.php
matthewhilton: Done

classes/local/store/azure_blob_storage/client.php
matthewhilton: Done

PR #654, opened 2024-12-11
deleter: check for any falsy value of deletelocal

SashaAnastasi:
Moodle stores the value `false` as an empty string in the database.
This results in undesired behaviour from `manipulator_can_execute()` since `'' == 0` evaluates to `true` in php7 but `false` in php8.


aspark21:
Should this change also be made to MOODLE_404_STABLE branch?
(for that matter should MOODLE_404_STABLE become the default branch?)

PR #662, opened 2025-01-23
Objectfs mimetype null/string handling

vithushau:
Fix for Task Failure in \tool_objectfs\task\push_objects_to_storage

The task \tool_objectfs\task\push_objects_to_storage was failing with the error:
`Return value must be of type string, null returned.`

The root cause was identified as files with an empty mimetype in the database, which were returning null instead of a valid string.

This PR introduces a fix to ensure that when an empty mimetype is encountered, an empty string ('') is returned instead of null. This aligns with the expected return type (string) and prevents the task from failing.

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -178,11 +178,12 @@ abstract class object_file_system extends \file_system_filedir {
 
         // We limit 1 because multiple files can have the same contenthash.
         // However, they all have the same mimetype so it does not matter which one we query.
-        return $DB->get_field_sql('SELECT mimetype
+        $mimetype = $DB->get_field_sql('SELECT mimetype
                               FROM {files}
                              WHERE contenthash = :hash
                              LIMIT 1',
                         ['hash' => $contenthash]);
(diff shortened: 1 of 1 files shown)


golenkovm:
Thanks for working on this @vithushau 

Please find some feedback below:
- Could you please update the PR description so it's more clear what kind of issue we had and how it's being fixed?
- Also, if you could take a look a see why the CI is upset that would be good. As far as I can see the last commits in both https://github.com/catalyst/moodle-tool_objectfs/commits/MOODLE_402_STABLE/ and https://github.com/catalyst/moodle-tool_objectfs/commits/MOODLE_404_STABLE/ were green
- Could you please also remove the ACL change, I'd like to tackle it separately.

Cheers,
Misha

golenkovm:
Thanks @vithushau 
It looks like in some cases `$DB->get_field_sql()` can return `false`. I believe `$mimetype ?? ''` may return `false` which would trigger the original error. Let's play safe and leverage `empty()` here.

PS: Thanks for fixing other bits.

PR #663, opened 2025-01-24
Objectfs mimetype null/string handling

vithushau:
Fix for Task Failure in \tool_objectfs\task\push_objects_to_storage

The task \tool_objectfs\task\push_objects_to_storage was failing with the error:
`Return value must be of type string, null returned.`

The root cause was identified as files with an empty mimetype in the database, which were returning null instead of a valid string.

This PR introduces a fix to ensure that when an empty mimetype is encountered, an empty string ('') is returned instead of null. This aligns with the expected return type (string) and prevents the task from failing.


golenkovm:
Thanks @vithushau 

PS: Same change for 4.2 branch: https://github.com/catalyst/moodle-tool_objectfs/pull/662

PR #666, opened 2025-02-13
Issue #596: SQL change with some performance / resource improvements

gbarat87:
This change would change the compatibility of this branch to Mysql 8+

#596


brendanheywood:
@gbarat87 is there a reason we need to bump the mysql verison? doing this on an old stable seems slightly risky to me

Issue #675, opened 2025-03-27
Creation of dynamic property $bucketacl is deprecated


leonstr:
With debugging enabled one sees on admin pages such as `admin/index.php?cache=1`:

_Deprecated: Creation of dynamic property tool_objectfs\local\store\s3\client::$bucketacl is deprecated in .../admin/tool/objectfs/classes/local/store/s3/client.php on line 81_

### Environment

Moodle 4.5.3+ (Build: 20250326)
tool_objectfs 2025031800
PHP version 8.2.27

PR #676, opened 2025-03-27
Fix Creation of property bucketacl deprecated #675

leonstr:
Fix warning PHP Deprecated: Creation of dynamic property tool_objectfs\local\store\s3\client::$bucketacl is deprecated.

Proposed fix for #675.


danmarsden:
Thanks @leonstr!

PR #702, opened 2025-08-26
Fix error alert on prod objectfs

duyhuynhdev:
Fix error alert on prod objectfs


classes/local/store/object_file_system.php
```diff
@@ -334,15 +336,12 @@ public function copy_object_from_external_to_local_by_hash($contenthash, $object
...
-            $localpath = $this->get_local_path_from_hash($contenthash);
             $localdirpath = $this->get_fulldir_from_hash($contenthash);
 
             // Folder may not exist yet if pulling a file that came from another environment.
-            if (!is_dir($localdirpath)) {
-                if (!mkdir($localdirpath, $this->dirpermissions, true)) {
-                    // Permission trouble.
-                    throw new file_exception('storedfilecannotcreatefiledirs');
-                }
+            if (!is_dir($localdirpath) && !@mkdir($localdirpath, $this->dirpermissions, true) && !is_dir($localdirpath)) {
```
bwalkerl: Do we really want to suppress mkdir errors? My gut feel is this should still logged, especially when an exception is thrown. Would also recommend leaving a comment as to why we have the second check.

Also do we know what is creating a directory between calls? Is this something making a directory without a lock or a lock not working? This looks correct to me, but I'm a bit concerned if there can be other conflicts later then the fix could potentially be worse than the current issue.

classes/local/store/object_file_system.php
```diff
@@ -334,15 +336,12 @@ public function copy_object_from_external_to_local_by_hash($contenthash, $object
 
         if ($initiallocation === OBJECT_LOCATION_EXTERNAL) {
 
-            $localpath = $this->get_local_path_from_hash($contenthash);
```
bwalkerl, line 337: While this is unused, I don't think it makes sense to clean this up alongside these fixes. Maybe as a separate commit.

bwalkerl (commented):
Thanks @duyhuynhdev 

This commit is fixing a handful of different issues, some of which are mostly unrelated. Can we please split this into separate commits with descriptive commit messages?

Something like:

- Ensure locks are released properly
- Add fallback check for directories created between calls

As we don't have an issue raised for these it would also be worth explaining why these are a problem.

classes/local/store/object_file_system.php
duyhuynhdev: @bwalkerl , I did some research and found this is the common practice for this scenario. If you search for `@mkdir`, you’ll see the `is_dir → @mkdir → is_dir` pattern used widely, including in core. That’s why I used @, though I agree it’s risky. On PHP < 8.0 it can [suppress critical errors](https://www.php.net/manual/en/language.operators.errorcontrol.php). So I will remove it in the next commit.

classes/local/store/object_file_system.php
duyhuynhdev: I’ll split this out and push it with the fallback check in the same commit.

classes/local/store/object_file_system.php
bwalkerl: More than anything the reason we want the warnings is because we don't know what else is creating them, and that's the biggest worry, especially since this function should have a lock.

classes/local/store/swift/client.php
```diff
@@ -90,8 +90,7 @@ private function get_endpoint() {
                     $openstack = new \OpenStack\OpenStack($endpoint);
                     $this->config->openstack_authtoken = $openstack->identityV3()->generateToken($endpoint)->export();
                     manager::set_objectfs_config(['openstack_authtoken' => serialize($this->config->openstack_authtoken)]);
-                    $lock->release();
-                } catch (\Exception $e) {
+                } finally {
                     $lock->release();
                 }
```
bwalkerl, line 94: This is a change in the behavior since exception goes form being caught to uncaught

classes/local/store/object_file_system.php
brendanheywood: get_contentdir_from_hash returns the comon hash prefix, so two processes could be working on 2 files with their own locks but by chance still have a common prefix and thats fine

brendanheywood:
Just muddying the waters slightly at this one got merged which fixes some of the same bits:

https://github.com/catalyst/moodle-tool_objectfs/pull/707

@duyhuynhdev will rebase this and remove the duplicate bits and we'll re-review shortly

duyhuynhdev:
> Just muddying the waters slightly at this one got merged which fixes some of the same bits:
> 
> #707
> 
> @duyhuynhdev will rebase this and remove the duplicate bits and we'll re-review shortly

It has been re based and waiting for a new review

renderer.php
```diff
@@ -213,13 +213,14 @@ private function get_output($fs, $url, $file, $identifier) {
...
+            '<a href="' . $url . '">' . $file->get_filename() . '</a>';
         if ('iframesnotsupported' === $identifier) {
-            $output = '<iframe height="400" width="100%" src="' . $url . '">'.
-                get_string('presignedurl_testing:' . $identifier, 'tool_objectfs').'</iframe>';
+            $output = '<iframe height="400" width="100%" src="' . $url . '">' .
+                get_string('presignedurl_testing:' . $identifier, 'tool_objectfs') . '</iframe>';
         }
-        return $output . '<br><small>' . $redirect . $url . '</small>';;
+        return $output . '<br><small>' . $redirect . $url . '</small>';
+        ;
```
bwalkerl: Remove duplicate ;

classes/local/store/object_file_system.php
```diff
@@ -632,10 +636,11 @@ public function xsendfile($contenthash) {
...
         }
         $headers = headers_list();
-        if ($this->presigned_url_configured() &&
+        if (
+            $this->presigned_url_configured() &&
                 $this->is_file_readable_externally_by_hash($contenthash) &&
-                $this->presigned_url_should_redirect($contenthash, $headers)) {
-
+                $this->presigned_url_should_redirect($contenthash, $headers)
+        ) {
```
bwalkerl, line 643: Since this now starts on a new line the following lines should have the same indenting

classes/local/store/object_file_system.php
```diff
@@ -959,8 +964,10 @@ public function presigned_url_should_redirect_file($file) {
      */
     public function presigned_url_should_redirect($contenthash, $headers = []) {
         // Redirect regardless.
-        if ($this->externalclient->presignedminfilesize == 0 &&
-                manager::all_extensions_whitelisted()) {
+        if (
+            $this->externalclient->presignedminfilesize == 0 &&
+                manager::all_extensions_whitelisted()
```
bwalkerl: Since this now starts on a new line the following lines can have the same indenting

classes/local/store/object_file_system.php
```diff
@@ -599,17 +601,19 @@ public function xsendfile_file(stored_file $file): bool {
         }
 
         $contenthash = $file->get_contenthash();
-        if ($this->presigned_url_configured() &&
+        if (
+            $this->presigned_url_configured() &&
                 $this->presigned_url_should_redirect_file($file) &&
-                $this->is_file_readable_externally_by_hash($contenthash)) {
-
+                $this->is_file_readable_externally_by_hash($contenthash)
```
bwalkerl: Since this now starts on a new line the following lines should have the same indenting

classes/local/store/object_file_system.php
```diff
@@ -599,17 +601,19 @@ public function xsendfile_file(stored_file $file): bool {
...
             return $this->redirect_to_presigned_url($contenthash, headers_list());
         }
 
         $ranges = $this->get_valid_http_ranges($file->get_filesize());
-        if ($this->externalclient->support_presigned_urls() && !empty($ranges) &&
-                $this->is_file_readable_externally_by_hash($contenthash)) {
-
+        if (
+            $this->externalclient->support_presigned_urls() && !empty($ranges) &&
+                $this->is_file_readable_externally_by_hash($contenthash)
```
bwalkerl: Since this now starts on a new line the following lines should can the same indenting

classes/local/store/object_file_system.php
```diff
@@ -180,11 +177,13 @@ protected function get_mimetype_from_hash(string $contenthash): string {
...
         // We limit 1 because multiple files can have the same contenthash.
         // However, they all have the same mimetype so it does not matter which one we query.
-        $mimetype = $DB->get_field_sql('SELECT mimetype
+        $mimetype = $DB->get_field_sql(
+            'SELECT mimetype
                               FROM {files}
                              WHERE contenthash = :hash
                              LIMIT 1',
-                        ['hash' => $contenthash]);
+            ['hash' => $contenthash]
```
bwalkerl, line 185: CI moving this to a new line has messed up indenting.

tests/local/object_manipulator/pusher_test.php
```diff
@@ -68,7 +67,8 @@ protected function set_pusher_config($key, $value) {
     public function test_pusher_get_candidate_objects_will_get_local_objects(): void {
         $object = $this->create_local_object();
 
-        self::assertTrue($this->objects_contain_hash($object->contenthash));;
+        self::assertTrue($this->objects_contain_hash($object->contenthash));
+        ;
```
bwalkerl: Remove duplicate ;

bwalkerl (commented):
Thanks @duyhuynhdev 

Based on @brendanheywood comments this looks good to merge. There's still a couple of small issues from the automatic CI fixes that should be cleaned up first.

Issue #698, opened 2025-08-04
Scheduled task failure (tool_objectfs\task\push_objects_to_storage)


ppv1979:
The log of scheduled task has the following error:

====================
Execute scheduled task: Object file system upload task (tool_objectfs\task\push_objects_to_storage)
... started 18:43:00. Current memory use 14.8 Мбайт.
... used 5 dbqueries
... used 0.13418507575989 seconds
Scheduled task failed: Object file system upload task (tool_objectfs\task\push_objects_to_storage),Coding error detected, it must be fixed by a programmer: A lock was created but not released at:
/var/www/devmoodle/moodle/admin/tool/objectfs/classes/local/store/object_file_system.php on line 317

 Code should look like:

 $factory = \core\lock\lock_config::get_lock_factory('type');
 $lock = $factory->get_lock(tool_objectfs_object_tool_objectfs: c0b5621f4beba1405d370ea4bdfca7c3847a33e8);
 $lock->release();  // Locks must ALWAYS be released like this.

Backtrace:
* line 86 of /admin/tool/objectfs/classes/local/object_manipulator/manipulator_builder.php: call to core\lock\lock->__destruct()
* line 69 of /admin/tool/objectfs/classes/task/task.php: call to tool_objectfs\local\object_manipulator\manipulator_builder->execute()
* line 410 of /lib/classes/cron.php: call to tool_objectfs\task\task->execute()
* line 194 of /admin/cli/scheduled_task.php: call to core\cron::run_inner_scheduled_task()
===================

Can you fix it?

Moodle 4.5.5+ (Build: 20250731)
tool_objectfs 2025040900

dragos5436:
Hi there,

Although this task does not fail for us anymore, we get the following error message whenever we create a new test Moodle instance:

<img width="701" height="626" alt="Image" src="https://github.com/user-attachments/assets/a3f636c1-b850-4cdb-bbd4-bdb9d124060c" />


Sometimes it might show up like this before logging in:

<img width="1096" height="818" alt="Image" src="https://github.com/user-attachments/assets/27d300b4-cc85-4651-a23a-7ffe8202c557" />

From looking at it, I can see it is largely the same error message as the one reported above.

The error is easy to get around, a refresh or navigation to a different path will make it go away and it won't come back for the same user. However, the first time a user will access the Moodle instance they will come across this error.

[...]

PR #707, opened 2025-10-20
Ensure Object Lock is released in case of exceptions

giovanne-oliveira:
Refactor object manipulation to ensure object lock is released after processing, even in case of exceptions.

Resolves issue #698


brendanheywood:
thanks @giovanne-oliveira :)

PR #670, opened 2025-03-17
Added Path Style Endpoint

Frostist:
Used for alternative software endpoints like MinIo.

I am wanting to add this setting to the stable branch. Which will add a tickbox to allow uses that use alternative S3 storage methods other than AWS to store there data. 

With Min.io it can't use the AWS standard and uses the "path style endpoint" which the AWS S3 plugin integrates with.

PR #716, opened 2026-01-12
S3 path style endpoint setting

Frostist:
I edited the code in a newer version of the repo in order to merge it now, see my past pull request here:
https://github.com/catalyst/moodle-tool_objectfs/pull/670


Frostist:
I have merged current updates here @danmarsden and @fDougelseSa

classes/local/store/s3/client.php
```diff
@@ -183,6 +183,10 @@ public function set_client($config) {
         if ($config->s3_base_url) {
             $options['endpoint'] = $config->s3_base_url;
         }
+        
+        if (!empty($config->s3_path_style_endpoint)) {
+            $options['use_path_style_endpoint'] = true;
+        }
```
natjw, line 189: This does not allow ti set it to false. What about:
```php
if (!empty($config->s3_path_style_endpoint)) {
    $options['use_path_style_endpoint'] = (bool) $config->s3_path_style_endpoint;
}
```

natjw (changes requested):
Like I said in my answer in my PR, there are 3 more things for this to work for me.

1. Is the thing mentioned in the review, the option is broken the way you hardcode it to `true`.
2. We need to catch another exception in case of error. See https://github.com/natjw/moodle-tool_objectfs/blob/74a69f015bfe6e565d531fd584bbdebef97b4fdc/classes/local/store/s3/client.php#L337-L340
3. We need to prevent the AWS SDK from using the IMS in case of not running on AWS. See https://github.com/natjw/moodle-tool_objectfs/blob/74a69f015bfe6e565d531fd584bbdebef97b4fdc/classes/local/store/s3/client.php#L821-L827

danmarsden:
closing this as #739 supercedes this - please test that and comment further there - (or submit a new PR with further changes on top!) :-)

Issue #718, opened 2026-01-16
Bug when searching config


brendanheywood:
we have settings such as 's3_bucket'

However this doesn't find them:

/admin/search.php?query=s3_bucket

<img width="352" height="123" alt="Image" src="https://github.com/user-attachments/assets/a62befe6-16fa-4f7a-bd31-374ff6de81d3" />

Need to figure out if this is a plugin bug or a core issue with the admin search

PR #719, opened 2026-01-23
Issue #718: Modified logic such that we can find objectfs settings via admin search.

abhinavgandham:
#718

## Summary

Fixed issue where specific client settings would not show up when searching for them in admin search.

**Before**
image

**After**
image

**NOTE**
Sorry the issue number in the branch is incorrect, it should be 718. Mistyped.

### DIFF
--- a/classes/local/store/object_client_base.php
+++ b/classes/local/store/object_client_base.php
@@ -112,6 +112,16 @@ abstract class object_client_base implements object_client {
     }
 
     /**
+     * Defines the setting section for the client. Logic is implemented by child classes.
+     *
+     * @param admin_settingpage $settings
+     * @param object $config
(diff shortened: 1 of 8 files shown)


classes/local/store/object_client_base.php
```diff
@@ -117,8 +117,12 @@ public function generate_presigned_url($contenthash, $headers = []) {
      * @return string
      * @throws /coding_exception
      */
-    public function define_client_check() {
+    public function define_client_check($objectfspage) {
         global $OUTPUT;
+        // Only going to perform the check if we are on the objectfs page.
+        if (!$objectfspage) {
```
brendanheywood: this should not be needed, the check is async via ajax and does not slow down the page having it loaded

PR #720, opened 2026-01-26
deleter: manipulator_can_execute check for any falsy deletelocal

SashaAnastasi:
Moodle stores the value false as an empty string in the database.
This results in undesired behaviour from `manipulator_can_execute()` since `'' == 0` evaluates to `true` in php7 but `false` in php8.


danmarsden:
this was already on an older branch but accidentally didn't land on the latest.

Issue #588, opened 2023-12-13
Extend file delete of local files to files which might have wrong location metadata


brendanheywood:
If you have a file which is marked as remote only, but the file is also still present locally then its metadata is wrong and it won't get picked up for file deletion.

If deletelocal is on then these should be removed, but if it is not then they should remain.

Even if deletelocal is on we should wait until most of the files have been correctly deleted before kicking in a second round of cleanup where we recurse filedir and look for any remaining files. The simplest thing it could do is just ensure that its location includes the bit for being local and then it will get deleted as per normal.

This could be the same task or another task, but it should not run unless deletelocal is on.

We previously tackled essentially this here but it did not land:

https://github.com/catalyst/moodle-tool_objectfs/issues/193

https://github.com/catalyst/moodle-tool_objectfs/pull/211/files

This will needs to be a task that can start and stop at any time and it will record a hash of the file it was last checking and processes all the dirs and subdirs in hash order. In normal operation I think this could run once a week and it should only run for a short while like an hour at most.

PR #724, opened 2026-02-22
[#588] Reconcile the filedir

alexdamsted:
This PR tries to address this issue: https://github.com/catalyst/moodle-tool_objectfs/issues/588

This PR performs two actions in one new scheduled task to address issue 588:
1. if the hash exists on disk, but not in the database then we should be safe to just delete the file
2. if the hash exists on disk and in mdl_files but no in objectfs then we add it with the metadata as local


classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,187 @@
...
+        // path = filedir/0a/bb/0abbcdef...
+        //
+        // This creates a two-level directory hierarchy:
+        //
+        // 1. Top-level directories: 00–ff (256 possible prefixes)
+        // 2. Second-level directories: 00–ff within each top-level directory
+        //
+        // Not all prefix combinations necessarily exist on disk and we don't care.
+        //
+        for ($i = 0; $i <= 255; $i++) {
```
brendanheywood: can you make $i and $j both start at whatever value $lasthash represents so we don't scan ones we don't need to scan

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,187 @@
...
+                // Sort the files to guarantee lexical hash order.
+                // Our resume logic and desired determinism requires some kind of order.
+                sort($files, SORT_STRING);
+
+                // Loop over each file.
+                foreach ($files as $filename) {
+                    // Quit after 1 hour.
+                    if ((time() - $starttime) >= $maxtime) {
+                        mtrace("ObjectFS: Time limit reached.");
+                        set_config('reconcile_file_lasthash', $lasthash, 'tool_objectfs');
```
brendanheywood, line 180: can you mtrace what hash it is finishing on

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,187 @@
...
+            mtrace('ObjectFS: deletelocal disabled — skipping filedir reconciliation.');
+            return;
+        }
+
+        // 1 hour max run time.
+        $starttime = time();
+        $maxtime = 3600;
+
+        // Get the last hash.
+        $lasthash = get_config('tool_objectfs', 'reconcile_file_lasthash');
```
brendanheywood, line 59: can you mtrace where it will start from

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,187 @@
...
+        //
+        // 1. Top-level directories: 00–ff (256 possible prefixes)
+        // 2. Second-level directories: 00–ff within each top-level directory
+        //
+        // Not all prefix combinations necessarily exist on disk and we don't care.
+        //
+        for ($i = 0; $i <= 255; $i++) {
+            $dir1 = sprintf('%02x', $i);
+
+            for ($j = 0; $j <= 255; $j++) {
```
brendanheywood: for the outer loop almost all of the 256 dirs will always exist, but for the inner loop very few will. Eg on a random stage site I just looked at most of these second level dirs only have dozen subdirs, and after we are done we may have even less. So I think instead of cycling through them all we should get a dir listing and only work through ones which exist, and which are a 2 digit hex code

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,187 @@
...
+
+                // Not all prefix combinations exist on disk.
+                if (!is_dir($subdir)) {
+                    continue;
+                }
+
+                // Scan the subdir for files.
+                $files = scandir($subdir);
+                if ($files === false) {
+                    continue;
```
brendanheywood, line 161: if there are no files I think we should delete the second level sub dir

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,243 @@
...
+                    // Persist progress frequently for safety.
+                    //
+                    // Every 100 files, we save our place in the traversal so the
+                    // task can resume safely without reprocessing or skipping
+                    // too many files.
+                    //
+                    // We do not want to set_config every loop, that would be slow.
+                    $processed++;
+                    if ($processed % 100 === 0) {
+                        set_config('reconcile_file_lasthash', $lasthash, 'tool_objectfs');
```
brendanheywood, line 275: can you add an mtrace here too please with overall $processed count and the hash

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,243 @@
...
+
+        // The config deletelocal must be enabled.
+        if (!get_config('tool_objectfs', 'deletelocal')) {
+            mtrace('ObjectFS: deletelocal disabled — skipping filedir reconciliation.');
+            return;
+        }
+
+        // 1 hour max run time.
+        $starttime = time();
+        $maxtime   = 3600;
```
brendanheywood: can you make this use tool_objectfs | maxtaskruntime

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,243 @@
...
+                    }
+
+                    // Now we get the current file's full path.
+                    $fullpath = $subdir . '/' . $filename;
+
+                    // Check if file exists in mdl_files.
+                    $fileexists = $DB->record_exists('files', ['contenthash' => $filename]);
+
+                    if (!$fileexists) {
+                        // Not referenced, safe to delete.
```
brendanheywood: can you put this behind a check for tool_objectfs | deletelocal

classes/task/reconcile_filedir.php
brendanheywood: also every 100 records can you also check this which allows the task to gracefully finish early if cron is trying to shut down:

```
crawler/classes/robot/crawler.php-553-            if (
crawler/classes/robot/crawler.php:554:                \core\local\cli\shutdown::should_gracefully_exit() ||
crawler/classes/robot/crawler.php-555-                \core\task\manager::static_caches_cleared_since($cronstart)
crawler/classes/robot/crawler.php-556-            ) {
c
```

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,243 @@
...
+                        if (!$objectexists) {
+                            $record = (object)[
+                                'contenthash'     => $filename,
+                                'location'        => OBJECT_LOCATION_LOCAL,
+                                'timeduplicated'  => 0,
+                                'filesize'        => filesize($fullpath) ?: null,
+                            ];
+                            $DB->insert_record('tool_objectfs_objects', $record);
+                            $inserted++;
+                        }
```
brendanheywood, line 260: just realized there is missing logic here, lets say that the tool_objectfs_objects knows about the hash, and thinks that the file is ONLY remote, we need to update its state so that it knows that it is both remote and local, and then the normal cron will run which will come along later and remove it from local if deletelocal is set.

Lets also track the number of $updated records as well.

If you insert or update, I think its better to use upsert_object(stdClass $object, $newlocation) {

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,274 @@
...
+                        mtrace("ObjectFS: Checkpoint at {$processed} files. Current hash {$lasthash}");
+
+                        // Allow graceful shutdown if cron is trying to exit.
+                        if (
+                            \core\local\cli\shutdown::should_gracefully_exit() ||
+                            \core\task\manager::static_caches_cleared_since($cronstart)
+                        ) {
+                            mtrace("ObjectFS: Graceful shutdown requested.");
+                            mtrace("ObjectFS: Finishing on hash {$lasthash}");
+                            mtrace("
```
brendanheywood: this will mtrace 4 lines, you want to break it into 4 strings which are concatted

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,274 @@
...
+
+                        mtrace("ObjectFS: Checkpoint at {$processed} files. Current hash {$lasthash}");
+
+                        // Allow graceful shutdown if cron is trying to exit.
+                        if (
+                            \core\local\cli\shutdown::should_gracefully_exit() ||
+                            \core\task\manager::static_caches_cleared_since($cronstart)
+                        ) {
+                            mtrace("ObjectFS: Graceful shutdown requested.");
+                            mtrace("ObjectFS: Finishing on hash {$lasthash}");
```
brendanheywood: probably redundant as you mtrace this on line 245

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
alexdamsted: Thanks. Added a new commit to address this.

classes/task/reconcile_filedir.php
```diff
@@ -0,0 +1,310 @@
...
+
+        // Tracking of processed, deleted, inserted and updated files.
+        $processed = 0;
+        $deleted   = 0;
+        $inserted  = 0;
+        $updated   = 0;
+        $skipped   = 0;
+
+        // Get the path to the filedir.
+        $filedir = $CFG->dataroot . '/filedir';
```
brendanheywood: use $CFG->filedir

alexdamsted: Thanks. Added a new commit to address this.

PR #729, opened 2026-03-11
Implemented new check class for cloud storage connection and added to settings page.

abhinavgandham:
Added a new check class that checks the status of the `test_connection` method used by other methods such as `define_client_section`. This has been added into the objectfs settings page.

### DIFF
--- a/classes/local/store/azure_blob_storage/client.php
+++ b/classes/local/store/azure_blob_storage/client.php
@@ -226,6 +226,28 @@ class client extends object_client_base {
     }
 
     /**
+     * Check if the client configured properly.
+     *
+     * @param \stdClass $config Client config.
+     * @return bool
+     */
+    public function is_configured($config) {
+        if (empty($config->azure_container)) {
+            return false;
+        }
+
+        if (empty($config->azure_accountname)) {
+            return false;
+        }
+
+        if (empty($config->azure_sastoken)) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
      * Tests connection
      * @return stdClass with 'success' and 'details' values.
      */
@@ -280,7 +302,7 @@ class client extends object_client_base {
         $settings->add(new \admin_setting_heading(
             'tool_objectfs/azure',
             new \lang_string('settings:azure:header', 'tool_objectfs'),
-            $this->define_client_check()
+            ''
         ));
 
         $settings->add(new \admin_setting_configtext(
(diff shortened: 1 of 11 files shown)


lang/en/tool_objectfs.php
```diff
@@ -37,12 +40,14 @@
...
 $string['check:tokenexpiry:na'] = 'Token expiry check not implemented for filesystem, or no token is set';
 $string['check_objects_location_task'] = 'Object file system check objects location task';
+$string['checkconnection'] = 'Object store connection status';
 $string['checkproxy_range_request'] = 'Pre-signed URL range request proxy';
 $string['checktagging_migration_status'] = 'Object tagging migration status';
 $string['checktagging_status'] = 'Object tagging';
 $string['checktagging_sync_status'] = 'Object tagging sync status';
 $string['checktoken_expiry'] = 'Token expiry';
 $string['client_not_available'] = 'The configured remote client is not available. Please ensure it is installed correctly.';
+$string['connectionstatus'] = 'Connection status';
```
brendanheywood, line 50: do we need 2 or 3 strings almost the same?

abhinavgandham: Yeah we can just combine these strings.

classes/local/store/azure_blob_storage/client.php
```diff
@@ -225,6 +225,28 @@ public function test_permissions($testdelete): stdClass {
         ];
     }
 
+    /**
+     * Check if the client configured properly.
+     *
+     * @param \stdClass $config Client config.
+     * @return bool
+     */
+    protected function is_configured($config) {
```
brendanheywood: public vs protected?

danmarsden:
Can we please make sure it doenst throw an error warning when the plugin only has read only access to the external storage?

lang/en/tool_objectfs.php
```diff
@@ -25,6 +25,9 @@
  */
 
 $string['backportfiletypesclass'] = 'Backport MDL-53240 is missing. Follow up https://github.com/catalyst/moodle-tool_objectfs#applying-core-patches';
+$string['check:connection:error'] = 'Could not establish connection to the object store. {$a}';
+$string['check:connection:na'] = 'No object store client is configured.';
```
alexdamsted: Small thing, the full stop looks good but isn't consistent with other check output strings. Should either remove or add full stops to the others.

abhinavgandham: Will just remove the full stops.

brendanheywood:
Dan all we have done here is move the existing checks to be done async so they don't delay all unrelated admin pages. So if that is actually the case lets make a new issue and address it there

PR #735, opened 2026-04-02
Add's large object support in swift

catalystfd:
This adds large object support to the swift filesystem.

Downloads are now streamed, resulting in downloads starting without waiting for buffering.
    
Range requests are now passed to swift, this improves seeking in video content significantly.

This requires the user use an updated local_openstack version with a more modern SDK, so the availability function check now validates that the openstack dependency has the correct version.


catalystfd:
I'll need to bump the version on this, but ideally #734 gets merged first.

catalystfd:
This is ready for review/merge

PR #743, opened 2026-06-11
742 remove external check

### LINKED ISSUE #742: Consider removing call to is_file_readable_externally_by_hash() from object_file_system::xsendfile_file
is_file_readable_externally_by_hash() results in a curl call to the repsective file storage. Eliminating is_file_readable_externally_by_hash() could result in performance gains.

jaypha:
Resolves #742 

Create a new function is_file_stored_externally_by_hash() that checks the external availability purely by local metadata.

Replaces calls to is_file_readable_externally_by_hash() with this function in xsendfile() and xsendfile_file().

By doing this, it avoids making calls to the remote server, this improving performance.

Also, some unrelated CI fixes are made in a separate commit.

### DIFF
--- a/version.php
+++ b/version.php
@@ -25,7 +25,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2026041000;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version   = 2026041001;      // The current plugin version (Date: YYYYMMDDXX).
 $plugin->release   = 2026041000;      // Same as version.
 $plugin->requires  = 2024042200;      // Requires 4.4.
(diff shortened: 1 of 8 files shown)


version.php
```diff
@@ -25,7 +25,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version   = 2026041000;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version   = 2026041001;      // The current plugin version (Date: YYYYMMDDXX).
 $plugin->release   = 2026041000;      // Same as version.
```
brendanheywood: can you bump the release in lock step please, everything else looks great

PR #747, opened 2026-06-17
Improve the reconcile task location detection

brendanheywood:


classes/task/reconcile_filedir.php
```diff
@@ -244,17 +277,34 @@ public function execute(): void {
                         );
 
                         if (!$object) {
+                            // If the filesystem is available, check whether the file already
+                            // exists remotely so we can register the correct location immediately
+                            // and avoid redundant processing by the push task.
+                            if ($filesystem !== null) {
+                                $location = $filesystem->get_object_location_from_hash($filename);
```
alexdamsted, line 284: This must be a lot slower right? A lot of requests, e.g., 1 per file via is_file_readable_externally_by_hash() -> is_readable() on the external file path

classes/task/reconcile_filedir.php
```diff
@@ -46,6 +68,17 @@ public function execute(): void {
             return;
         }
 
+        // Instantiate the filesystem once so we can check remote existence without
+        // making a new connection per file.
```
alexdamsted, line 72: is this true? it looks like there's one call per file on like 284 at $location = $filesystem->get_object_location_from_hash($filename);

classes/task/reconcile_filedir.php
alexdamsted: this happens in another task anyway, we shift the processing time to the reconcile filedir task in this case which makes sense.

Issue #750, opened 2026-07-01
Add configurable exclusions for presigned URL redirects by component and file area


Anupamatd:
ObjectFS redirects eligible Moodle file requests to presigned URLs when presigned URL redirects are enabled. Some Moodle components, however, load files using XHR/fetch from within an iframe or other contexts where redirecting to a presigned URL can cause CORS preflight failures.

Introduce an admin setting that allows administrators to configure which Moodle component/filearea pairs should bypass presigned URL redirects.

**Background**:
SCORM packages that use XHR (XMLHttpRequest) to load content files (audio, CSS, JS, JSON, images) fail with a CORS error when objectfs presigned URL redirects are enabled.

**Steps to reproduce:**
* Configure objectfs with S3/CloudFront and enable presigned URL redirects
* Upload a SCORM package created with Adobe Captivate (that uses XHR-based resource loading)
* Launch the SCORM activity
* Observe that media files (e.g., MP3 audio) fail to load

**Expected behaviour:**
All SCORM content files load successfully within the SCORM player.

**Actual behaviour:**
The browser console shows:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource ...... (Reason: CORS preflight response did not succeed). Status code: 403
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at ...... (Reason: CORS request did not succeed). Status code: (null)
```

PR #751, opened 2026-07-01
Issue #750: Add configurable exclusions for presigned URL redirects by component and file area

Anupamatd:
Add configurable exclusions for presigned URL redirects by component and file area


classes/local/store/object_file_system.php
```diff
@@ -877,6 +877,13 @@ public function presigned_url_should_redirect_file($file) {
             return false;
         }
 
+        // Avoid presigned URL redirects for SCORM content.
+        // SCORM packages often load resources via XHR/fetch from an iframe, and
+        // cross-origin redirects can cause browser CORS failures.
+        if ($file->get_component() === 'mod_scorm' && $file->get_filearea() === 'content') {
```
brendanheywood: can you change this to a generic admin setting which is a text area, each line would be an area eg

```
mod_scorm|content
```

Set that as the default value, and then if we find more we can add them on the fly

Anupamatd: Thank you @brendanheywood - Implemented an admin setting to configure which Moodle component|filearea pairs should bypass presigned URL redirects. Please review. Thanks

classes/local/store/object_file_system.php
```diff
@@ -877,12 +877,60 @@ public function presigned_url_should_redirect_file($file) {
...
             manager::is_extension_whitelisted($file->get_filename()));
     }
 
+    /**
+     * Returns true if this file belongs to a component/filearea pair configured to avoid redirects.
+     *
+     * @param stored_file $file
+     * @return bool
+     */
+    private function is_disallowed_presigned_filearea(stored_file $file): bool {
```
brendanheywood: add unit tests please

Anupamatd: Thank you @brendanheywood - I've added a unit test. Please review it and let me know if there’s anything else I need to cover. Thanks

PR #753, opened 2026-07-03
Issue #752: Migrate all the location status to bits

### LINKED ISSUE #752: Migrate all the location status to bits
- [ ] 1 = in mdl_files
- [ ] 2 = in filedir
- [ ] 3 = in remote store

eg change all these:

https://github.com/catalyst/moodle-tool_objectfs/blob/d0e75304361ddb0d6a12a85ec473390b37c917f6/lib.php#L43-L61

petersistrom:
Resolves #752 

**Schema: boolean columns instead of a bitmask column**

Moodle has no cross-database support for bitwise indexes, so a single combined location integer column cannot be indexed efficiently. Replacing it with three boolean columns (in_filedir, in_mdl_files, in_remote) allows a standard composite index to work on all supported databases.

**PHP/DB boundary**

PHP code keeps the existing bitmask constants. location_helper translates at the boundary: splitting a bitmask into the three boolean values on write and reconstructing it from a DB row on read.

**Candidate classes**

The candidates/ directory and its abstraction layer have been removed. Each manipulator now owns its candidate query directly as a static get_candidates(stdClass $config): array method.


lib.php
```diff
@@ -29,36 +29,62 @@
...
+define('OBJECT_LOCATION_MISSING', 2);
 
 /**
- * Location enum of the object
- * LOCAL is when the object exists locally only.
+ * Location of the object: in filedir and mdl_files, not yet pushed to any remote store.
+ * Value: OBJECT_LOCATION_IN_FILEDIR | OBJECT_LOCATION_IN_MDL_FILES = 3.
  */
-define('OBJECT_LOCATION_LOCAL', 0);
+define('OBJECT_LOCATION_LOCAL', 3);
```
brendanheywood: can we actually define these in terms of the other defines?

petersistrom: Updated

db/install.xml
```diff
@@ -9,7 +9,7 @@
         <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
         <FIELD NAME="contenthash" TYPE="char" LENGTH="40" NOTNULL="true" SEQUENCE="false"/>
         <FIELD NAME="timeduplicated" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="false"/>
-        <FIELD NAME="location" TYPE="int" LENGTH="1" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="location" TYPE="int" LENGTH="3" NOTNULL="true" SEQUENCE="false"/>
```
alexdamsted: upgrade.php will need to update the precision, e.g., line 208 in upgrade.php. 

Do something like this in the new upgrade.php 2026041006 step:

```
    $table = new xmldb_table('tool_objectfs_objects');
    $field = new xmldb_field(
        'location', XMLDB_TYPE_INTEGER, '3', null,
        XMLDB_NOTNULL, null, null, 'timeduplicated'
    );
    $dbman->change_field_precision($table, $field);
```

classes/log/aggregate_logger.php
```diff
@@ -65,6 +65,8 @@ public function __construct() {
         parent::__construct();
         $this->movestatistics = [
             OBJECT_LOCATION_ERROR => [],
```
alexdamsted: `OBJECT_LOCATION_ERROR (0)` can now no longer be returned, since you now do this: `define('OBJECT_LOCATION_IN_MDL_FILES', 2);`

Every scheduled task run will log an empty 'error' location bucket,
adding noise. The error location bucket should be removed.

classes/log/aggregate_logger.php
```diff
@@ -159,14 +161,16 @@ public function location_to_string($location) {
         switch ($location) {
             case OBJECT_LOCATION_ERROR:
```
alexdamsted: This is dead code now too

classes/local/store/object_file_system.php
```diff
@@ -708,6 +709,7 @@ public function delete_object_from_hash($contenthash) {
 
         switch ($location) {
             case OBJECT_LOCATION_LOCAL:
+            case OBJECT_LOCATION_ORPHANED:
```
alexdamsted: Is OBJECT_LOCATION_ORPHANED unreachable?

delete_object_from_hash() only ever gets values from
get_object_location_from_hash() of which always ORs in IN_MDL_FILES (bit 2).

A value of 1 (ORPHANED = IN_FILEDIR only, no IN_MDL_FILES) can never be
returned by the function.

classes/local/report/object_status_history_table.php
```diff
@@ -240,11 +240,11 @@ public function get_file_location_string($filelocation) {
             'filedir' => 'object_status:filedir',
             'deltaa' => 'object_status:delta:a',
             'deltab' => 'object_status:delta:b',
-            OBJECT_LOCATION_ERROR => 'object_status:location:error',
+            OBJECT_LOCATION_MISSING => 'object_status:location:error',
```
alexdamsted: should the lang string be object_status:location:missing ?

classes/local/store/object_file_system.php
```diff
@@ -288,20 +288,21 @@ public function is_file_readable_externally_by_hash($contenthash) {
...
-            return OBJECT_LOCATION_DUPLICATED;
-        } else if ($localreadable && !$externalreadable) {
-            return OBJECT_LOCATION_LOCAL;
-        } else if (!$localreadable && $externalreadable) {
-            return OBJECT_LOCATION_EXTERNAL;
-        } else {
-            // Object is not anywhere - we toggle an error state in the DB.
-            manager::update_object_by_hash($contenthash, OBJECT_LOCATION_ERROR);
-            return OBJECT_LOCATION_ERROR;
+        // The mdl_files bit is always set because this function is only called
```
alexdamsted: Is the condition this comment asserts actually enforced?

E.g., If the function were ever called for a contenthash not present in mdl_files it would return LOCAL (3) instead of ORPHANED (1) right?

db/install.xml
petersistrom: Updated upgrade.php to include the field change

classes/log/aggregate_logger.php
petersistrom: Removed

classes/log/aggregate_logger.php
petersistrom: Removed

classes/local/report/object_status_history_table.php
petersistrom: Updated the lang string

classes/local/store/object_file_system.php
petersistrom: Removed

classes/local/store/object_file_system.php
petersistrom: Yes correct, if it were called for a file not in mdl_files it would return LOCAL.  The precondition isn't enforced, it's just a calling convention. However, orphan detection is handled via scheduled tasks.

classes/local/report/object_status_history_table.php
alexdamsted: > Updated the lang string

Where is it updated? OBJECT_LOCATION_MISSING still refers to the lang string 'object_status:location:error' but I'm thinking it should refer to a lang string named 'object_status:location:missing' to make more sense.

petersistrom: Apologies, didn't make it in to the commit initially. Updated now

classes/local/store/object_file_system.php
brendanheywood: The recoverer task looks for objects which are in the objectfs table but not in mdl_files and then tries to recover  them. I'd prefer it if this had an param which if set did proper checks to see if its in the mdl_files as well.

Or perhaps better, it would pass in a bit map of the places we know it is present and then we don't have to re-check those places. So most places start with knowing mdl_files is present, the recoverer on the other hand only knows that it is present in objectfs tables.

get_object_location_from_hash($contenthash, $knownlocations = OBJECT_LOCATION_IN_MDL_FILES) {


recoverer:
get_object_location_from_hash($contenthash, 0) {

petersistrom: Yes, I agree passing in a bit map of known locations is a better approach. Happy to implement that.

petersistrom: Updated to pass in known locations

classes/local/store/object_file_system.php
```diff
@@ -306,24 +306,35 @@ public function is_file_readable_externally_by_hash($contenthash) {
...
+     *                            Defaults to OBJECT_LOCATION_IN_MDL_FILES since most callers reach
+     *                            this function via a stored_file or mdl_files reference.
+     *                            Pass 0 when the caller only knows the object is in the objectfs
+     *                            table (e.g. the recoverer), so all bits are checked fresh.
      * @return int
      */
-    public function get_object_location_from_hash($contenthash) {
-        $localreadable = $this->is_file_readable_locally_by_hash($contenthash);
-        $externalreadable = $this->is_file_readable_externally_by_hash($contenthash);
+    public function get_object_location_from_hash($contenthash, $knownlocations = OBJECT_LOCATION_IN_MDL_FILES) {
```
alexdamsted: The default for $knownlocations should be 0 and callers of this method should pass OBJECT_LOCATION_IN_MDL_FILES (add this to the doc block of this method to make it known to callers).

We want to do this because if a caller calls this on a hash in the db table objectfs_objects but the associated file reference of the hash no longer exists in mdl_files then the bitmask returned will incorrectly include IN_MDL_FILES which causes the checker to class the object as EXTERNAL or DUPLICATED instead of MISSING or the like.

alexdamsted:
@petersistrom Can the deleter, puller, pusher, and recoverer candidates classes now just be deleted? They aren't reachable via the factory anymore from your changes.

@brendanheywood mentioned about the need of candidates now after your changes. To me looking at the code raw it looks like a layer of abstraction that just complicates things more. You could move the sql from these into a single candidates static class to simplify the abstraction perhaps. 

What problem was the candidates abstraction solving before your changes? To me the candidates had different sql logic and has some useful unique methods, e.g., relating to logging, but now with your changes its not as much the case and the unique methods aren't a lot of code.

petersistrom:
@alexdamsted Yes they can be deleted now, they were providing different SQL queries for each task but that has now collapsed into one query with the bitmask changes. And I think Brendan's suggestion now makes sense to nuke the concept of candidate classes completely.

classes/local/store/object_file_system.php
petersistrom: Yeah I agree that's much safer, updated to pass in 0 as the default

PR #758, opened 2026-07-22
Normalise pre-signed URLs pasted into editors #368

### LINKED ISSUE #368: Re normalize cut and pasted signed urls back into moodle plugin files

### LINKED ISSUE #744: Rewrite signed urls back to plugin urls on text area saves
Will require a new hook in core:

https://moodle.atlassian.net/browse/MDL-73596

See also:

https://github.com/catalyst/moodle-filter_presignedurl/issues/4

bwalkerl:
Requires [MDL-73596](https://moodle.atlassian.net/browse/MDL-73596) to be backported to make use of this feature.

Testing instructions:

1. Setup objectfs and enable pre-signing
2. Open a pluginfile in a new page, you should see a full pre-signed URL. With this patch it should have a `#objectfs-origin=` at the end. 
3. Open an editor. This can be mod_label (Text and media area) or anything else. 
4. Copy the URL and paste it into an editor. If this is an image, the image will be shown. If not, you will see the full URL. Hyperlinks can also be tested. 
5. Click save.
6. Open the editor and confirm that you see a pluginfile link instead. For images, confirm this by viewing the source code in the editor.

This will only work if the link has character encoding for spaces, which should be the default for copying from URL in browser. Chromium will always have the encoded URL, while Firefox will decode the URL but have users copy the encoded URL unless it has been modified. In such edge cases where the URL has been modified (or copied from a different source) the copied value can contain spaces instead of the encoded value, which will break the current handling (it _should_ be obvious in the editor this is broken saving, so I don't think this is the main concern).

Closes #368 
Closes #744

### DIFF
--- a/classes/local/store/object_file_system.php
+++ b/classes/local/store/object_file_system.php
@@ -53,6 +53,9 @@ require_once($CFG->libdir . '/filestorage/file_storage.php');
  * [Description object_file_system]
  */
 abstract class object_file_system extends \file_system_filedir {
+    /** @var string Origin for pre-signed urls. */
+    public const PLUGINFILE_ORIGIN_SEGMENT_PREFIX = 'objectfs-origin=';
+
     /**
      * @var mixed
      */
@@ -888,6 +891,30 @@ abstract class object_file_system extends \file_system_filedir {
         }
     }
 
+    /**
+     * Convert pre-signed URLs inside text back to pluginfiles.
+     *
+     * @param string $text The content that may contain URLs in need of rewriting.
+     * @return string The processed text.
+     */
+    public function normalise_presigned_urls(string $text): string {
+        if (!$this->presigned_url_configured()) {
+            return $text;
+        }
+
+        $re = sprintf(
+            '!https?://[^\s<>"\']*?\#%s([^\s<>"\']+\w)!',
+            preg_quote(self::PLUGINFILE_ORIGIN_SEGMENT_PREFIX)
+        );
+
+        // The origin URL should only have minimal RFC 3986 encoding on anchors.
+        // These characters should be rare and even desirable to keep encoded in pluginfile links.
+        return preg_replace_callback(
+            $re,
+            fn($matches) => (new \moodle_url($matches[1]))->out(),
+            $text
+        );
+    }
 
     /**
      * Return if the file system supports presigned_urls.
--- a/tests/object_file_system_test.php
+++ b/tests/object_file_system_test.php
@@ -18,6 +18,7 @@ namespace tool_objectfs;
 
 use coding_exception;
 use tool_objectfs\local\store\object_file_system;
+use tool_objectfs\local\store\signed_url;
 use tool_objectfs\local\manager;
 use tool_objectfs\local\tag\tag_manager;
 use tool_objectfs\tests\test_client;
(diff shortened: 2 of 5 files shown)


tests/object_file_system_test.php
```diff
@@ -1269,4 +1269,113 @@ public function test_is_file_stored_externally_by_hash(): void {
...
+
+        $embeddedurls = [];
+        foreach ($pluginfiles as $index => $pluginfile) {
+            $ME = $pluginfile;
+            $embeddedurls[] = $filesystem->embed_origin_url($presignedurl);
+
+            // Verify the origin is added as a fragment.
+            $fragment = (string)parse_url($embeddedurls[$index], PHP_URL_FRAGMENT);
+            $this->assertStringContainsString(object_file_system::PLUGINFILE_ORIGIN_SEGMENT_PREFIX, $fragment);
+            $this->assertStringContainsString(urlencode($pluginfile), $fragment);
```
alexdamsted: I think rawurlencode should be used for urls?

classes/local/store/object_file_system.php
```diff
@@ -880,13 +883,59 @@ public function redirect_to_presigned_url($contenthash, $headers = []) {
...
+        if (empty($me)) {
+            debugging(OBJECTFS_PLUGIN_NAME . ': Missing origin for pre-signed url.', DEBUG_DEVELOPER);
+            return $presignedurl;
+        }
+
+        return sprintf(
+            "%s#%s%s",
+            $presignedurl,
+            self::PLUGINFILE_ORIGIN_SEGMENT_PREFIX,
+            urlencode($me)
```
alexdamsted: I think rawurlencode should be used again here?

classes/local/store/object_file_system.php
```diff
@@ -880,13 +883,59 @@ public function redirect_to_presigned_url($contenthash, $headers = []) {
...
+     */
+    public function embed_origin_url($presignedurl): string {
+        $me = me();
+        if (empty($me)) {
+            debugging(OBJECTFS_PLUGIN_NAME . ': Missing origin for pre-signed url.', DEBUG_DEVELOPER);
+            return $presignedurl;
+        }
+
+        return sprintf(
+            "%s#%s%s",
```
alexdamsted: What if the presigned url already has a fragment? E.g., you could get something like #fragment#objectfs-origin=...

Maybe a guard for that is good here to strip the existing fragment.

classes/local/store/object_file_system.php
```diff
@@ -880,13 +883,59 @@ public function redirect_to_presigned_url($contenthash, $headers = []) {
...
+     * @param string $presignedurl
+     * @return string presignedurl with embedded origin
+     */
+    public function embed_origin_url($presignedurl): string {
+        $me = me();
+        if (empty($me)) {
+            debugging(OBJECTFS_PLUGIN_NAME . ': Missing origin for pre-signed url.', DEBUG_DEVELOPER);
+            return $presignedurl;
+        }
+
```
alexdamsted: Before returning here, if $me contains a space we should add some debugging to at least log that a space is detected and the aforementioned spacing issue is happening. This is the earliest time we can detect a space so I think the debugging should be here.

classes/local/store/object_file_system.php
bwalkerl: I'll have to verify what a $ME can contain. This was part of the previous PR, so I'm not sure whether raw url encoding is actually required or if this is just for spaces.

classes/local/store/object_file_system.php
bwalkerl: Strictly speaking, the fragment is just so it can be appended to the URL and not intended to actually work as a fragment. A presigned URL should not come with an existing fragment, and on the off chance they do multiple fragments are covered in unit tests. That being said, there may be a case to use fragments properly.

classes/local/store/object_file_system.php
bwalkerl: The mentioned spacing issue is only for HTML parsing in the editor (if for some reason this is copied from somewhere other than the URL bar or somewhere that doesn't convert spaces). It shouldn't be part of $ME.

classes/local/store/object_file_system.php
alexdamsted: Its a minor thing but it would encode spaces as %20, seems more semantically correct imo

classes/local/store/object_file_system.php
alexdamsted: Thanks. The debugging could be placed in the normalise_presigned_urls function, after the replacement? Somewhere just to log if this issue occurs.

classes/local/store/object_file_system.php
bwalkerl: I've refactored this and let moodle_url handle the encoding of the fragment.

classes/local/store/object_file_system.php
bwalkerl: Now that this uses `moodle_url->set_anchor()` is replaces the existing fragment.

classes/local/store/object_file_system.php
bwalkerl: I don't think this is necessary since the conditions to trigger it are mostly known (although they may differ per browser), and the output wouldn't provide any new information. I'm still open to improving the detection in the regex in the text though.

bwalkerl: I've looked into this a bit more and noticed that the edge case of of copying a URL after it has been modified is firefox only, so would be rarer than I initially thought. I've updated the PR description to describe this, but ultimately that's a user error.

tests/object_file_system_test.php
```diff
@@ -1269,4 +1270,122 @@ public function test_is_file_stored_externally_by_hash(): void {
...
+
+    /**
+     * Test rewriting pre-signed URLs to pluginfiles.
+     *
+     * @dataProvider normalise_presigned_urls_provider
+     * @covers ::normalise_presigned_urls
+     *
+     * @param array $config
+     */
+    public function test_normalise_presigned_urls(array $config): void {
```
brendanheywood, line 1403: I would like to see some test examples with real signed urls, which would have a redacted domain and also be stale shortly anyway, and make sure they get mapped back to the right plugin file.

PR #762, opened 2026-08-05
Use check API for proxy range request test #618

### LINKED ISSUE #543: The pre-signed URL check triggers on every admin page

### LINKED ISSUE #618: Move objectfs checks to check API classes, include checks in admin settings
Since Moodle 4.4 https://tracker.moodle.org/browse/MDL-67898 is integrated, allowing you to show check API results https://moodledev.io/docs/4.4/apis/subsystems/check in the admin settings very easily.

Currently objectfs rolls this itself, for example:

https://github.com/catalyst/moodle-tool_objectfs/blob/c4ba44f1fe9a85c20c76a423169a9f0ce771edc2/settings.php#L162-L165

The reason for rolling itself is probably only due to the fact that objectfs is older than the check API.

Will help resolve #543

bwalkerl:
Closes #618 
Closes #543


petersistrom:
Hi @bwalkerl 
Your changes in settings.php look good, however I did find some issues with the proxy range request itself, maybe a a separate issue is required? But thought I'd flag it here:

The two returns below are misleading, the check is executed even if `proxyrangerequestsdisabled` is disabled, for testing this PR when the connection test failed it returns `Unknown The proxy range request setting is disabled.` which is not the case.

https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_404_STABLE/classes/check/proxy_range_request.php#L55

https://github.com/catalyst/moodle-tool_objectfs/blob/MOODLE_404_STABLE/classes/check/proxy_range_request.php#L71

bwalkerl:
@petersistrom That's a good point. 

I think the intention here was to always run the test to know if enabling it will work, so from that view the test is currently working the way we want, and the check is only added to the status page if the setting is enabled.  But the messages are certainly misleading. 

I'll update the messages as part of this.

petersistrom:
Thanks @bwalkerl, that makes sense. Happy to merge this one in.

PR #763, opened 2026-08-05
Disallow presigned urls based on cache headers #469

### LINKED ISSUE #469: Have an option to not sign cache control immutable / public files
Lets say we have a theme file which is not baked into the theme source but it is served through the pluginfile api. Eg I have a prod theme file which serves a file like this:

/pluginfile.php/1/theme_xxxx/banner/576336053/connections-abstract-fuschia-micro-banner.jpg

The pluginfile function sends some headers for public immutable and long expires exactly as it should. 

cache-control: public, max-age=5184000, no-transform

* [x] Proposing a admin setting which when on, allows public immutable urls to not be signed and let them be served normally by php and then cached at the CDN layer.

* [x] Have an option for a list of urls (components + file areas) which should be public but which are not, and override their caching headers

bwalkerl:
Test instructions:
1. Setup objectfs and enable Pre-Signed URLs.
2. Upload a file with https://github.com/catalyst/moodle-tool_publicfiles or find any other immutable public files. 
3. Start by enabling the setting and confirm Pre-Signing is used on the file (partly to confirm Pre-Signing is working)
4. Disable the new setting 
5. Open dev tools and disable caching as the Pre-Signed URL will still likely be cached. 
6. With dev tools still open load the file and confirm the file isn't redirected to the Pre-Signed URL.  

Closes #469


classes/local/store/object_file_system.php
```diff
@@ -975,6 +981,49 @@ public function is_disallowed_presigned_filearea(stored_file $file): bool {
...
+        }
+
+        // Immutable assets are long-lived by definition.
+        if (in_array('immutable', $directives)) {
+            return true;
+        }
+
+        // Also treat high max-age values as long-lived, even without immutable.
+        foreach ($directives as $directive) {
+            if (str_contains($directive, 'max-age')) {
```
petersistrom: This would also catch `s-maxage` so if both are present in the headers, it will take the first potentially incorrect value

bwalkerl: Right now it only catches max-age, but good call that should also be handled. Did you see any places in Moodle that set that?

settings.php
```diff
@@ -289,6 +289,13 @@
                 PARAM_RAW_TRIMMED
             ));
 
+            $settings->add(new admin_setting_configcheckbox(
+                'tool_objectfs/disallowlonglivedcache',
```
brendanheywood: lets call this presignpublicurls and it should default to off

classes/local/store/object_file_system.php
```diff
@@ -975,6 +981,57 @@ public function is_disallowed_presigned_filearea(stored_file $file): bool {
...
+    /**
+     * Returns true when cache headers represent a disallowed cache policy.
+     *
+     * This can be used to avoid redirecting to pre-signed URLs when Moodle has
+     * already emitted response headers suitable for long-term CDN/browser caching.
+     *
+     * @param array $headers Request headers.
+     * @return bool
+     */
+    public function is_disallowed_presigned_cache_policy(array $headers): bool {
```
brendanheywood: is_disallowed_presigned_cache_policy I think could be renamed to is_url_publicly_cacheable

tests/object_file_system_test.php
```diff
@@ -797,6 +797,61 @@ public function test_is_disallowed_presigned_filearea_refreshes_lookup_when_conf
...
+     * Tests detection and handling of cache control headers.
+     *
+     * @dataProvider is_disallowed_presigned_cache_policy_provider
+     * @covers ::is_disallowed_presigned_cache_policy
+     * @param array $headers response headers.
+     * @param bool $expected expected result.
+     * @return void
+     */
+    public function test_is_disallowed_presigned_cache_policy(array $headers, bool $expected): void {
+        set_config('disallowlonglivedcache', 0, 'tool_objectfs');
```
brendanheywood: the admin setting does not match the code config item

classes/local/store/object_file_system.php
bwalkerl: I updated it to `should_presign_cache_headers` so that the setting can be checked inside of it.

PR #764, opened 2026-08-19
Refactor location checks to bitwise logic and add extensibility hooks

petersistrom:


db/hooks.php
```diff
@@ -0,0 +1,33 @@
...
+ * @package   tool_objectfs
+ * @copyright Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+$callbacks = [
+    // Intentionally empty: tool_objectfs defines and dispatches hooks but does not listen to them itself.
+];
```
bwalkerl: Is this blank file needed to dispatch hooks?

classes/local/manager.php
```diff
@@ -224,6 +227,12 @@ public static function upsert_object(stdClass $object, $newlocation) {
             $fs->push_object_tags($object->contenthash);
         }
 
+        // Dispatch hook for multi-store placement tracking.
+        if ($locationchanged) {
+            $hook = new \tool_objectfs\hook\after_object_upsert($object, $newlocation, $oldlocation);
+            \core\di::get(\core\hook\manager::class)->dispatch($hook);
+        }
```
bwalkerl, line 234: The hook name implies it fires after all upserts, but it's only fired when location changed. Perhaps the name needs to be updated?

classes/local/store/object_file_system.php
```diff
@@ -224,11 +225,35 @@ protected function get_remote_path_from_hash($contenthash) {
...
+            \core\di::get(\core\hook\manager::class)->dispatch($hook);
+            if ($hook->is_resolved()) {
+                $resolved = $hook->get_resolved_path();
+                if (is_string($resolved) && $resolved !== '') {
+                    return $resolved;
+                }
+            }
+        }
+
+        return $path;
```
bwalkerl: Is true an appropriate default for check readable? This could be a performance concern for remote files.

classes/hook/get_additional_location_bits.php
```diff
@@ -0,0 +1,83 @@
...
+    /** @var array Additional bit => column mappings registered by listeners. */
+    private array $additionalbits = [];
+
+    /**
+     * Register an additional store bit column.
+     *
+     * @param int $bit The bit value (must be a power of 2, >= 8).
+     * @param string $column The DB column name on tool_objectfs_objects.
+     */
+    public function add_bit(int $bit, string $column): void {
```
bwalkerl, line 42: Should the column name be validated?

bwalkerl (commented):
Code is looking mostly good to me, just a couple of concerns

db/hooks.php
petersistrom: No. Moodle's hook system only requires db/hooks.php in plugins that listen to hooks (register callbacks). I've deleted this file

classes/local/manager.php
petersistrom: I've renamed the hook class to after_object_location_changed and updated the 4 files that referenced it.

classes/local/store/object_file_system.php
petersistrom: Good catch, I've updated the default to false so no network round-trip on every path resolution.

classes/hook/get_additional_location_bits.php
petersistrom: Added validation to the new column names

classes/local/store/object_file_system.php
bwalkerl: Looks good. Now my question is whether there's any code in objectfs where it should be true.

petersistrom: oh yeah good call, have updated two places 
https://github.com/catalyst/moodle-tool_objectfs/blob/bitwise-hooks/classes/local/store/object_file_system.php#L219 - the file isn't local, so we're about to serve from external. If the primary store is unreachable, the hook can resolve an alternative.

https://github.com/catalyst/moodle-tool_objectfs/blob/bitwise-hooks/classes/local/store/object_file_system.php#L871 -  need a readable source to copy from. If the primary store is down, fallback to another store.

bwalkerl: I think those are still relatively hot paths? If so, maybe this should be locked behind a setting?

petersistrom: I've gated the external path check behind a config. Might be worth improving error/retry flow in the future so the config can be removed.

bwalkerl: Honestly even the full config on this path may still be a bit too much, but should be better than before. Lets just get this merged for now and can improve on it later.

PR #765, opened 2026-08-25
Clean up settings page

bwalkerl:
- Re-add s3 credential provider chain check that was removed by mistake in #729 -  checks whether the it is safe to enable the use credential provider chain setting. This has been converted to use the check API and only provides information relevant to the use of a single setting so is only required on settings page when not checked. The results of this check have been also updated to INFO and lang strings made more descriptive.
- Move connection status to the above where the client settings are set, so status is visible when updating client settings.
- Cherry-pick #641

### DIFF
--- a/classes/check/s3_sdkcreds.php
+++ b/classes/check/s3_sdkcreds.php
@@ -1,0 +1,63 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_objectfs\check;
+
+use core\check\check;
+use core\check\result;
+use tool_objectfs\local\manager;
+
+/**
+ * Status check for S3 SDK Credentials.
+ *
+ * @package    tool_objectfs
+ * @author     Benjamin Walker <benjaminwalker@catalyst-au.net>
+ * @copyright  Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class s3_sdkcreds extends check {
+    /**
+     * Link to ObjectFS settings page.
+     *
+     * @return \action_link|null
+     */
+    public function get_action_link(): ?\action_link {
+        $url = new \moodle_url('/admin/category.php', ['category' => 'tool_objectfs']);
+        return new \action_link($url, get_string('pluginname', 'tool_objectfs'));
+    }
+
+    /**
+     * Get the result of the connection check using sdk credentials.
+     *
+     * @return result
+     */
+    public function get_result(): result {
+        $config = manager::get_objectfs_config();
+        $config->s3_usesdkcreds = 1;
+        $client = manager::get_client($config);
+
+        if (empty($client) || !$client->is_configured($config)) {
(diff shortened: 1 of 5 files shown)


classes/check/s3_sdkcreds.php
```diff
@@ -0,0 +1,63 @@
...
+        $config->s3_usesdkcreds = 1;
+        $client = manager::get_client($config);
+
+        if (empty($client) || !$client->is_configured($config)) {
+            return new result(result::NA, get_string('check:connection:na', 'tool_objectfs'));
+        }
+
+        $connection = $client->test_connection();
+        if ($connection->success) {
+            return new result(result::INFO, get_string('settings:aws:sdkcredsok', 'tool_objectfs'));
```
petersistrom, line 58: This check returns result::INFO for both success and failure, which makes the status output indistinguishable and inconsistent with the other checks in this plugin e.g. https://github.com/catalyst/moodle-tool_objectfs/blob/ef7572aa8005d16f998bba975c3bfafdff2e10df/classes/check/connection.php#L58

lang/en/tool_objectfs.php
```diff
@@ -143,8 +143,8 @@
 $string['settings:aws:key_prefix_help'] = 'Prefix to use inside Amazon S3 bucket. Must end with trailing slash when set. Leave blank to use root of bucket.';
 $string['settings:aws:region'] = 'region';
 $string['settings:aws:region_help'] = 'Amazon S3 API gateway region.';
-$string['settings:aws:sdkcredserror'] = 'Couldn\'t find AWS credentials. It\'s unsafe to enable this setting. Follow up <a href="https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html">AWS documentation</a>.';
-$string['settings:aws:sdkcredsok'] = 'AWS credentials found. This setting can be safely enabled.';
+$string['settings:aws:sdkcredserror'] = 'AWS credentials not found in the credential provider chain. Using the chain without credentials is unsafe, use an S3 key and secret instead. Follow up <a href="https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html">AWS documentation</a>.';
```
petersistrom, line 146: get_string() sends $connection->details as a variable but is not used in the string https://github.com/catalyst/moodle-tool_objectfs/blob/b522f403490b0a0a922650a260c80069be3b664e/classes/check/s3_sdkcreds.php#L61

settings.php
```diff
@@ -207,6 +214,8 @@
         \tool_objectfs\local\manager::get_available_fs_list()
     ));
 
+    $settings->add(new admin_setting_check('tool_objectfs/connectionstatus', new connection(), true));
```
petersistrom, line 217: admin_setting_check only exists in 4.5+ should be surrounded with a if class_exists similar to https://github.com/catalyst/moodle-tool_objectfs/blob/ef7572aa8005d16f998bba975c3bfafdff2e10df/settings.php#L342

petersistrom:
Code looks good, just a couple of comments above

settings.php
bwalkerl: Despite one  comment suggesting otherwise I think it should be in 4.4. Looks to have landed in MDL-67898. We have other checks not using that.

lang/en/tool_objectfs.php
bwalkerl: I copied that from the connection check but you're right it's not needed as we don't use it here and it's not used on the status page. I'll remove it.

classes/check/s3_sdkcreds.php
bwalkerl: Some extra context:
- Previously, this message was in the description for enabling use sdkcreds. 
- I had a request to change the previous warning to info because it was causing some confusion as it's completely safe to use a key and secret instead. 

I initially left OK as is, but it became apparent that similarly it's not really representing an OK state. This check is for when it's safe to enable use sdkcreds, but a green notification in this location could be mistaken for the connection working - which wouldn't be true unless that setting is enabled. This was slightly less of a concern before when it was on the description of another setting and not it's own check. 

As such I think info is most appropriate for both states, even if it doesn't stand out as much.

petersistrom: Makes sense happy to them both as info

